PR #2, opened 2026-01-27
moxygen::vanilla API design / plan

afrind:


michalhosna:
I'd like to clarify the goal here. Is the intent to:
- Use this new API in the relay itself?
- Upstream it to moxygen?
- Provide an integration path for non-relay consumers?

I generally support `std` over `folly` where reasonable. I see two categories of changes here:

**1. Equivalent replacements**

Changes like `folly::Optional` → `std::optional` seem like no-brainers; the differences are minimal as far as my (and Claude's) understanding goes. 

Are there any substantial advantages to `folly::Optional` over `std`? Wouldn't it make sense to propose switching to `std::optional` in upstream moxygen as well?

**2. Lossy replacements**

Replacing `IOBuf` with copy-into-vector seems like a significant downgrade. Minimizing copies and deferring consolidation into contiguous memory is an important performance optimization in network applications, especially when operating in QUIC, where buffers are already in userspace within the same process.

I would rather not see trade-offs of this kind in the relay.

afrind:
> I'd like to clarify the goal here. Is the intent to:
> 
> * Use this new API in the relay itself?

My initial thought is no, but I'm open to discussion about it.  The main difference is using coroutines in the control plane, and I think once folks boot into this world they will find it much easier to reason about.

> * Upstream it to moxygen?

I've mentioned it to the rest of the Meta team and I think they are not averse to holding it upstream - which would mean it would get updated with e.g. API changes they make. 

> * Provide an integration path for non-relay consumers?

I think this is the primary goal.  As such, it's not the highest pri for relay from my perspective.

> Changes like `folly::Optional` → `std::optional` seem like no-brainers; the differences are minimal as far as my (and Claude's) understanding goes.
> 
> Are there any substantial advantages to `folly::Optional` over `std`? Wouldn't it make sense to propose switching to `std::optional` in upstream moxygen as well?

folly::Optional predates std::optional and I believe has one minor API difference.  I think a moved-from std::optional is undefined, but a moved from folly::Optional is the empty optional (eg: std::nullopt).

The std::optional change would absolutely be in upstream.

> **2. Lossy replacements**
> 
> Replacing `IOBuf` with copy-into-vector seems like a significant downgrade. Minimizing copies and deferring consolidation into contiguous memory is an important performance optimization in network applications, especially when operating in QUIC, where buffers are already in userspace within the same process.

I think claude offered a way to make an IOBuf backed Payload from a Payload constructed with eg: an owned vector 

```
      // Transfer vector ownership to IOBuf via custom deleter
      auto* data = vec.data();
      auto size = vec.size();
      auto* vecPtr = new std::vector<uint8_t>(std::move(vec));
      return folly::IOBuf::takeOwnership(
          data, size,
          [](void*, void* userData) {
            delete static_cast<std::vector<uint8_t>*>(userData);
          },
          vecPtr);
```

> I would rather not see trade-offs of this kind in the relay.

100%

design/moxygen-vanilla-plan.md
```diff
@@ -0,0 +1,1657 @@
...
+struct FullTrackName { TrackNamespace trackNamespace; std::string trackName; /* operators... */ };
+
+// Message types (after std::optional migration)
+struct SubscribeRequest { ... };
+struct SubscribeOk { ... };
+struct SubscribeDone { ... };
+struct SubscribeUpdate { ... };
+struct Fetch { ... };
+struct FetchOk { ... };
+struct Announce { ... };
```
akash-a-n, line 104: Should we migrate to moq-15 spec names like PublishNamespace instead of the older Announce?
Not just here, I see references to AnnounceHandle.. etc.

michalhosna:
> > - Use this new API in the relay itself?
>
> My initial thought is no, but I'm open to discussion about it. The main difference is using coroutines in the control plane, and I think once folks boot into this world they will find it much easier to reason about.

I'd generally want a more "pure" interface based on std. However, to my knowledge there is no good (i.e. usable) standard API for coroutines or `IOBuf`-like functionality. I do not think switching away from folly for the relay is not a good trade-off right now. Splitting folly into different targets shaves off part of my worries here (thanks, Alan!).

If we wanted to get rid of `folly` down the line, we could always roll our API structures at that point.

> > - Provide an integration path for non-relay consumers?
>
> I think this is the primary goal. As such, it's not the highest pri for relay from my perspective.

Agreed on priority. Given that I don't plan to be a 3rd-party consumer, nor will I be seeing this in the relay, I am going to leave this PR review to others.

suhasHere:
Having a std api (not folly/co-routine centric) api is useful even for integrating openmoq relay component into server infrastructures. Coroutines encourages bring your own runtime and its not always a good option when dealing with lot of them in progress. I am fine with having an option of non-coroutines and std::c++ api as a priority ( i think we agreed on it at some point too )

design/moxygen-vanilla-plan.md
```diff
@@ -0,0 +1,1657 @@
...
+
+## Seven-Phase Approach
+
+### Phase 1: Migrate MoQFramer to std::optional
+First commit: Replace `folly::Optional` with `std::optional` in MoQFramer.h and related files.
+
+### Phase 2: Extract Core Types
+Second commit: Split core types out of MoQFramer.h into a new `Types.h` that both moxygen and moxygen::vanilla can depend on.
+
+### Phase 3: Replace IOBuf in Extensions
```
afrind, line 14: I changed my mind about this.  We'll leave both Payload and Extensions using folly::IOBuf and there can just be wrappers for consuming applications that need to get either spans() or a coalesced buffer.

On the publish side, we can add ingress wrappers as well.

design/moxygen-vanilla-plan.md
```diff
@@ -0,0 +1,1657 @@
...
+├── IOBufPayload.cpp     # IOBufPayload implementation + payload conversion
+├── Callbacks.h          # All callback interface definitions
+├── Publisher.h          # Publisher interface with callbacks
+├── Subscriber.h         # Subscriber interface with callbacks
+└── MoQConsumers.h       # TrackConsumer, SubgroupConsumer, FetchConsumer
+```
+
+---
+
+## 1. Expected.h - Custom Expected Type
```
afrind, line 335: std::expected was introduced in C++-23.  I don't think there's much point in creating a new wrapper since the concept is coming.  We can create moxygen::expected which either aliases to std::expected or mvfst has a std::expected compatible implementation.

afrind (commented):
Note Phases 1, 2 and 4 have landed in moxygen upstream already.  Phase 5 (executors) is in-progress.

design/moxygen-vanilla-plan.md
afrind: Yes, this is half done now (subNS and pubNS have replaced subAnnounces and announce).  There's another diff coming (likely today) fixing subDone -> pubDone.

afrind:
I'm merging this for now to clear out some old PRs.  Most of the preliminary steps have landed, and Suhas as a variation of the vanilla idea in #14.

We can always update/delete later.

PR #3, opened 2026-01-27
Module loader design

afrind:
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/3)

### DIFF
--- a/design/module-loader.md
+++ b/design/module-loader.md
@@ -1,0 +1,520 @@
+# Module Loader Design Plan
+
+## Overview
+
+A relay server plugin system using dynamic loading (`.so` files via `dlopen`/`dlsym`). Supports multiple plugin types with different interfaces: authentication, miss handlers, routing, etc.
+
+## Core Concepts
+
+### Plugin Types
+
+Each plugin type has:
+- A well-known interface (pure virtual class)
+- A factory function symbol name
+- Version compatibility info
+
+```cpp
+enum class PluginType {
+  Auth,
+  MissHandler,
+  Router,
+  // extensible...
+};
+```
+
+### Plugin Interface Base
+
+All plugins inherit from a common base for lifecycle management:
+
+```cpp
+class Plugin {
+ public:
+  virtual ~Plugin() = default;
+  virtual const char* name() const = 0;
+  virtual uint32_t version() const = 0;
+};
+```
+
+### Type-Specific Interfaces
+
+Each plugin type defines its interface:
+
+```cpp
+// Auth plugin interface
+class AuthPlugin : public Plugin {
+ public:
+  struct AuthResult {
+    bool allowed;
+    std::string reason;
+    // additional metadata...
+  };
+
+  virtual AuthResult authenticate(
+      const std::string& token,
+      const std::string& resource) = 0;
+};
+
+// Miss handler plugin interface
+class MissHandlerPlugin : public Plugin {
+ public:
+  virtual Task<FetchResult> fetch(FetchRequest req) = 0;
+};
+
+// Router plugin interface
+class RouterPlugin : public Plugin {
+ public:
+  virtual std::vector<Endpoint> route(const TrackName& track) = 0;
+};
+```
+
+## Module Loader
+
+```cpp
+class ModuleLoader {
+ public:
+  // Load a plugin from a shared object file, registered by name
+  // Returns nullptr on failure, logs error
+  template <typename T>
+  std::shared_ptr<T> load(const std::string& name, const std::string& path);
+
+  // Get a previously loaded plugin by name
+  template <typename T>
+  std::shared_ptr<T> get(const std::string& name);
+
+  // Get all loaded plugins of a type
+  template <typename T>
+  std::vector<std::shared_ptr<T>> getAll();
+
+  // Unload all plugins (call before shutdown)
+  void unloadAll();
+
+ private:
+  struct LoadedModule {
+    void* handle;           // dlopen handle
+    std::string path;
+    std::string name;       // config-specified name
+    std::shared_ptr<Plugin> instance;
+  };
+  std::vector<LoadedModule> modules_;
+};
+```
+
+Multiple plugins of the same type can be loaded. Config determines which to use:
+
+```cpp
+// Load multiple auth plugins
+loader.load<AuthPlugin>("ldap", "/usr/lib/relay/auth_ldap.so");
+loader.load<AuthPlugin>("jwt", "/usr/lib/relay/auth_jwt.so");
+loader.load<AuthPlugin>("anonymous", "/usr/lib/relay/auth_anon.so");
+
+// Config specifies which to use per scenario
+// e.g., "internal" namespace uses ldap, "public" uses jwt
+auto plugin = loader.get<AuthPlugin>(config.authPluginFor(request.namespace));
+```
+
+### Factory Function Convention
+
+Each plugin `.so` exports a C factory function. The function names include the plugin type for version checking per-type:
+
+```cpp
+// In an auth plugin .so:
+extern "C" {
+  AuthPlugin* create_auth_plugin(PluginHost* host);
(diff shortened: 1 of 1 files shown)


design/module-loader.md
```diff
@@ -0,0 +1,520 @@
...
+  int auth_plugin_authenticate(
+      auth_plugin_handle h,
+      const char* token,
+      const char* resource,
+      char* reason_out,
+      size_t reason_size);
+}
+```
+
+**Decision**: For moxygen, C++ returns are acceptable since plugins will be built with the same toolchain. If third-party plugins with arbitrary compilers become a requirement, a pure C adapter layer can be added later.
```
michalhosna: I am not sure if this is the right decision. 

Given that this will be an OSS project and we want to ship pre-built binaries (see [spec 2.1.8])(https://github.com/OpenMOQ/relay-requirements/blob/f9e11d6ebfc6298458b7a3f8f554b0e4ec9a6785/relay-requirements.md#21-core-technology-stack). 

I would expect that people would like to ship pre-built 3rd party modules. For the adoption of this project, I think it is beneficial to be able to ship pre-compiled modules. 

Sure, this can be solved by requiring a specific toolchain, but I am not sure if that's the greatest solution?

Having a stable ABI is, I think, beneficial for the adoption of this project. 

I really do not like pure C APIs, but here they seem appropriate not only for ABI stability, but also for the ability to provide modules in different (non-C++) languages. 

Having a C++ -> C "adapter" for both sides (relay and modules) should be able to remove the C pain here fairly easily, so I think we should go with it right away and avoid a API split down the line.

michalhosna (changes requested):

design/module-loader.md
afrind: Ok that sounds safer.  Sorry the AI was like "Decision: this is fine" lol.

Given the sophistication of some of the modules we're talking about (eg: black box, caching), that's a lot of C++ <-> C translation.  Does this decision mean we want to try to scope down the interfaces to something more simplistic that's easier to express in C?

michalhosna: I'd consider the C-API just a well-understood method to define a stable ABI, not the greatest one, but the universal one. It doesn't need to be the target for human programmers. We should avoid abominations in the C layer, but simplicity there should not be a focus. I would still push the C++ wrapper API as the primary interface

michalhosna (commented):
> @michalhosna reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/3)* status: all files reviewed, 1 unresolved discussion (waiting on afrind).

design/module-loader.md
afrind: I think we should do as you say and plumb the modules via C, but it's a pain.  Let me update the PR.

afrind (commented):
> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/3)* status: all files reviewed, 1 unresolved discussion (waiting on michalhosna).

michalhosna (changes requested):
> @michalhosna made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/3)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on afrind).

___
*[`design/module-loader.md` line 432 at r2](https://reviewable.io/reviews/openmoq/o-rly/3#-Omv2HT7B0QdixQLrugY:-Omv2HT7B0QdixQLrugZ:b-6b3z94) ([raw file](https://github.com/openmoq/o-rly/blob/a6aee43baac0c67c2d1e66f8320a6427f7da12a0/design/module-loader.md#L432)):*
> ```Markdown
> **Current approach (C++ returns)**: Simpler code, works when plugins are built with the same toolchain as the relay.
> 
> **Alternative (pure C interface)**: More portable across compilers. Would use opaque handles and C function pointers:
> ```

I think the whole "C++ vs Pure C Interface" section needs little bit of updating, if nothing else to not confuse AI later.

TimEvens (changes requested):
> @TimEvens made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/3)* status: 0 of 1 files reviewed, 7 unresolved discussions (waiting on afrind and michalhosna).

___
*[`design/module-loader.md` line 33 at r3](https://reviewable.io/reviews/openmoq/moqx/3#-OqBxXee8eBs3jFis6hj:-OqBxXee8eBs3jFis6hk:b-10pgjg) ([raw file](https://github.com/openmoq/moqx/blob/994b5ccc6b2f0b7d84f7d401be6cf64afce4219a/design/module-loader.md#L33)):*
> ```Markdown
>  public:
>   virtual ~Plugin() = default;
>   virtual const char* name() const = 0;
> ```

Prefer these be part of the constructor(s)
___
*[`design/module-loader.md` line 34 at r3](https://reviewable.io/reviews/openmoq/moqx/3#-OqBxkTk-ctjrLV_d3rI:-OqBxkTk-ctjrLV_d3rJ:b-6lfvdk) ([raw file](https://github.com/openmoq/moqx/blob/994b5ccc6b2f0b7d84f7d401be6cf64afce4219a/design/module-loader.md#L34)):*
> ```Markdown
>   virtual ~Plugin() = default;
>   virtual const char* name() const = 0;
>   virtual uint32_t version() const = 0;
> ```

Constructor is better to enforce expected variables.  

32bit is fine but likely will be encoded in dot notation such as **SemVer**. Might be better add string overload for version number to Set. We should have a Get() as well. 
___
*[`design/module-loader.md` line 40 at r3](https://reviewable.io/reviews/openmoq/moqx/3#-OqByaKe-4-Q5bXno1DY:-OqByaKe-4-Q5bXno1DZ:bpe96fx) ([raw file](https://github.com/openmoq/moqx/blob/994b5ccc6b2f0b7d84f7d401be6cf64afce4219a/design/module-loader.md#L40)):*
> ```Markdown
> ### Type-Specific Interfaces
> 
> Each plugin type defines its interface:
> ```

I feel that we will need to define additional abstract classes/interfaces for each plugin type.  For example, type `Auth` it is expected that there is `authenticate(token, resource)` method. 
___
*[`design/module-loader.md` line 78 at r3](https://reviewable.io/reviews/openmoq/moqx/3#-OqC-nr0D076vWCmyQN3:-OqC-nr0D076vWCmyQN4:b-zbp1yd) ([raw file](https://github.com/openmoq/moqx/blob/994b5ccc6b2f0b7d84f7d401be6cf64afce4219a/design/module-loader.md#L78)):*
> ```Markdown
>   // Returns nullptr on failure, logs error
>   template <typename T>
>   std::shared_ptr<T> load(const std::string& name, const std::string& path);
> ```

We need better documentation on params and methods.  Not just for this method, but for all abstract/interface classes. 

Let's use doxygen syntax, which is well supported by IDEs.  
___
*[`design/module-loader.md` line 82 at r3](https://reviewable.io/reviews/openmoq/moqx/3#-OqC0T-XDCKNF1vFIp3t:-OqC0T-XDCKNF1vFIp3u:b-jks7ex) ([raw file](https://github.com/openmoq/moqx/blob/994b5ccc6b2f0b7d84f7d401be6cf64afce4219a/design/module-loader.md#L82)):*
> ```Markdown
>   // Get a previously loaded plugin by name
>   template <typename T>
>   std::shared_ptr<T> get(const std::string& name);
> ```

should this be virtual?
___
*[`design/module-loader.md` line 86 at r3](https://reviewable.io/reviews/openmoq/moqx/3#-OqC0XVZ8Kvw6ohnXZ9y:-OqC0XVZ8Kvw6ohnXZ9z:b-8xo19j) ([raw file](https://github.com/openmoq/moqx/blob/994b5ccc6b2f0b7d84f7d401be6cf64afce4219a/design/module-loader.md#L86)):*
> ```Markdown
>   // Get all loaded plugins of a type
>   template <typename T>
>   std::vector<std::shared_ptr<T>> getAll();
> ```

virtual?

PR #5, opened 2026-01-27
Deduplication Design

afrind:


akash-a-n:
Hi @afrind ,
I am not sure if I missed this, I am trying to understand the necessity for deduplication. 
If we detect that we have multiple publishers for the same track, can we not unsubscribe from the duplicates? Is it beneficial to have multiple publishers somehow?

afrind:
Merging to clear the review queues.

PR #7, opened 2026-01-28
Update platform support section in module-loader.md

mondain:
Clarified platform support details for the plugin system, noting macOS limitations regarding multiple .so files and dependencies.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/7)

### DIFF
--- a/design/module-loader.md
+++ b/design/module-loader.md
@@ -421,7 +421,7 @@ plugins:
 
 ### Platform Support
 
-This plugin system uses `dlopen`/`dlsym` and works on systems that support these APIs (Linux, macOS, other POSIX systems). Windows is not currently supported.
+This plugin system uses `dlopen`/`dlsym` and works on systems that support these APIs (Linux, macOS, other POSIX systems). Windows is not currently supported. Multiple `.so` or those that have dependencies on other `.so` is disallowed on macOS; this will require special handling when macOS is the target OS.
 
 ### C++ vs Pure C Interface
 
--- a/design/module-loader.md
+++ b/design/module-loader.md
@@ -421,7 +421,7 @@ plugins:
 
 ### Platform Support
 
-This plugin system uses `dlopen`/`dlsym` and works on systems that support these APIs (Linux, macOS, other POSIX systems). Windows is not currently supported. Multiple `.so` or those that have dependencies on other `.so` is disallowed on macOS; this will require special handling when macOS is the target OS.
+This plugin system uses `dlopen`/`dlsym` and works on Linux. Windows and macOS builds will not support plugins initially.
 
 ### C++ vs Pure C Interface
 


mondain:
I couldn't find the comment that actually triggered the thought, but this section can contain this information.

design/module-loader.md
```diff
@@ -421,7 +421,7 @@ plugins:
 
 ### Platform Support
 
-This plugin system uses `dlopen`/`dlsym` and works on systems that support these APIs (Linux, macOS, other POSIX systems). Windows is not currently supported.
+This plugin system uses `dlopen`/`dlsym` and works on systems that support these APIs (Linux, macOS, other POSIX systems). Windows is not currently supported. Multiple `.so` or those that have dependencies on other `.so` is disallowed on macOS; this will require special handling when macOS is the target OS.
```
michalhosna: This is a great note, but I am not sure about the goal here. Are we pushing the solution down the line? 

I think we should either
1. Only do Linux support (per [the spec](https://github.com/OpenMOQ/relay-requirements/blob/f9e11d6ebfc6298458b7a3f8f554b0e4ec9a6785/relay-requirements.md#L87))
2. or do a proper non-linux support at every step. 

I would push for linux-only support. Generally there will be more places where those differences between linux/posix/bsd/macos pop-up. Doing abstractions above OS differences is sometimes a lot of work and often hard not to lose advantages of one of the systems (e.g. io_uring). 

I do not think it makes sense to support macOS just because developers use it (including me). Will anybody run this server-grade relay on macOS? I would not think so. It would make more sense to me to support BSD if somebody needs that; see https://github.com/OpenMOQ/relay-requirements/issues/8. 

I am happy to help people with VMs/containers dev setup on macOS. 

I am not going to block non-linux in principle alone, but we need to put the work there. I do not really have experience with non-Linux high performance server stuff. Can somebody describe how the Mac solution is supposed to work here? What's the special handling required? The rest of the documents talks about loading multiple .so's, how is this supposed to work on macos? 

-----

Sorry to bikeshed on a short PR, but I would like to resolve this question generically, and this short PR seems like a fitting place.

michalhosna (changes requested):

design/module-loader.md
afrind: I think we can say mac support is a low-pri side goal, and features like custom module loading might be self-supported.

mondain: Yes, I understood this as Linux only, but if we also support local builds for quick dev testing, the option of a macos build would be a nice-to-have for the 90% of people on the Team using macos (I am not one of them, but I understand the need).

mondain (commented):
> @mondain made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/7)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on michalhosna and suhasHere).

mondain (commented):
:lgtm:

> @mondain made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/7#-:-OoN9T22CE8_ThBNixfv:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on michalhosna and suhasHere).

design/module-loader.md
michalhosna: I am sorry, but I don't feel my arguments are addressed in the response.

michalhosna (changes requested):
> @michalhosna reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/7)* status: all files reviewed, 2 unresolved discussions (waiting on afrind, mondain, and suhasHere).

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/openmoq/o-rly/7#-Oo56cVh3uwLUF2-mNl9:-Oo56cVh3uwLUF2-mNlA:b2nh8oq):*
???

design/module-loader.md
mondain: Lets just remove it; the purpose of this was an informational note; it's not important enough to keep IMHO.

mondain (commented):
> @mondain made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/7)* status: all files reviewed, 3 unresolved discussions (waiting on afrind, michalhosna, and suhasHere).

___
*[`design/module-loader.md` line 424 at r1](https://reviewable.io/reviews/openmoq/o-rly/7#-Oo_MdD7A4bkw6xaf6hN:-Oo_MdD7A4bkw6xaf6hO:btt1nsh) ([raw file](https://github.com/openmoq/o-rly/blob/03048f7c1478c3cf5c826ec8be0c6a191368ca27/design/module-loader.md#L424)):*
> ```Markdown
> ### Platform Support
> 
> This plugin system uses `dlopen`/`dlsym` and works on systems that support these APIs (Linux, macOS, other POSIX systems). Windows is not currently supported. Multiple `.so` or those that have dependencies on other `.so` is disallowed on macOS; this will require special handling when macOS is the target OS.
> ```

Is this how I edit here?



_Suggestion:_
```Markdown
This plugin system uses `dlopen`/`dlsym` and works on target OS, which is Linux based.
```

mondain (commented):
> @mondain made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/7)* status: all files reviewed, 3 unresolved discussions (waiting on afrind, michalhosna, and suhasHere).

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/openmoq/o-rly/7#-Oo56cVh3uwLUF2-mNl9:-Oo_N3X5-XVVWv1Lan6A:bjtyqsq):*

???
</blockquote></details>

I don't understand

design/module-loader.md
afrind:
```suggestion
This plugin system uses `dlopen`/`dlsym` and works on Linux. Windows and macOS builds will not support plugins initially.
```

PR #8, opened 2026-01-31
Relay Project Layout

mondain:
Introduces the initial build and documentation scaffold for the OpenMOQ relay reference implementation. Adds C++20/CMake structure with scripts, Docker Ubuntu 20.04 baseline for glibc compatibility, and a minimal relay binary + smoke test to keep the build healthy. Includes clang-format/clang-tidy enforcement, CMake presets, and developer docs plus initial architecture and MOQT draft 14–16 roadmap.


docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:20.04
```
michalhosna: Why go with Ubuntu and not Debian directly? 

Especially given that it's customary in official Docker images to have Debian (i.e. not Ubuntu) and Alpine variants for server-like software. 

For example
- [nginx](https://hub.docker.com/_/nginx)
- [postgres](https://hub.docker.com/_/postgres) 
- [httpd](https://hub.docker.com/_/httpd)

design/ARCHITECTURE.md
```diff
@@ -0,0 +1,69 @@
...
+- Version gates should be explicit and easy to reason about.
+- Avoid undefined behavior and non-standard extensions.
+
+## Directory mapping (planned)
+
+- `src/transport/` QUIC bindings.
+- `src/protocol/` MOQT framing and session.
+- `src/relay/` route graph, cache, policies.
+- `src/store/` in-memory cache, optional disk.
+- `src/obs/` logging/metrics.
```
michalhosna: I don't think this document reflects the current consensus in https://github.com/OpenMOQ/relay-requirements/blob/main/relay-requirements.md

Generally, this document seems out of scope here. There are separate design PRs with more to come that are doing this in a more profound way.  Typically, I think single-purpose PRs are preferred so they are easier to review and quicker to merge.

design/ROADMAP.md
```diff
@@ -0,0 +1,59 @@
...
+- **Conformance testing**: develop a small scripted test harness for CI.
+- **Compatibility**: keep draft-specific logic isolated and well documented.
+- **Security**: input validation, limits, and denial-of-service safeguards.
+- **Documentation**: example config and operational guidance for operators.
+
+## Draft transition policy (expected)
+
+- Each draft is implemented behind explicit gates.
+- Default behavior targets the latest supported draft.
+- Backward compatibility should be tested, not assumed.
```
michalhosna: Same as above, this seems out-of-scope and should be much better covered by e.g. issues. 

I do understand such documents are useful for AI/Claude. But this seems too high level for anything we want to do and should be better served by the desing PRs

.gitignore
```diff
@@ -0,0 +1,55 @@
...
+# Logs
+*.log
+
+# Sanitizers/profilers
+*.profraw
+*.profdata
+
+# Generated binaries (local)
+o_rly
+o_rly_smoke
```
michalhosna: .gitignore should only contain project-specific ignores;

Having a long .gitignore makes it much easier to make mistakes and ignore something unintentionally. 

Ignoring `Testing/` directory anywhere in the tree seems dangerous. 

`o_rly` should be ignored by a full path, if at all; shouldn't that end up in some build folder? 

IDE specific or OS specific ignores should be handled by local configuration, not in-repo. Do we accept random PR for  .gitignore every time new IDEs pops up? 

See other OSS projects:
- https://github.com/nginx/nginx/blob/master/.gitignore

README.md
```diff
@@ -1,2 +1,11 @@
 # o-rly
-The OpenMOQ Relay
+The OpenMOQ Relay (reference implementation).
```
michalhosna:
```suggestion
The OpenMOQ Relay
```

Are we doing reference? 

I think we are clearly doing something much more and referencing the implementation of MOQT Relay.

.gitattributes
```diff
@@ -0,0 +1,13 @@
...
+*.h text eol=lf
+*.hpp text eol=lf
+*.json text eol=lf
+*.properties text eol=lf
+*.xml text eol=lf
+*.bat text eol=crlf
+*.sh text eol=lf
+*.txt text
+*.vcproj text eol=crlf
+
```
michalhosna: Why do we need this file? What's the benefit of having this in the repository and having to maintain it?

scripts/lint.sh
```diff
@@ -0,0 +1,19 @@
...
+set -euo pipefail
+
+BUILD_DIR=${1:-build}
+
+if [[ ! -f "${BUILD_DIR}/compile_commands.json" ]]; then
+  echo "compile_commands.json not found in ${BUILD_DIR}. Run configure first." >&2
+  exit 1
+fi
+
+FILES=$(rg --files -g '*.{h,hpp,cc,cpp,cxx}' include src tests tools)
```
michalhosna: Why use [`rg`](https://github.com/BurntSushi/ripgrep)? This seems like a use case that can be covered by plain POSIX `find`. And I am not sure requiring ripgrep makes sense at this point.

scripts/docker-build.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+IMAGE=${1:-o-rly-ubuntu20}
+
+docker build -t "${IMAGE}" -f docker/Dockerfile .
```
michalhosna: Ignorable nitpick:

```suggestion
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"

IMAGE=${1:-o-rly-ubuntu20}

docker build -t "${IMAGE}" -f "$PROJECT_ROOT"/docker/Dockerfile "$PROJECT_ROOT"
```

The `.` is relative to the caller's directory. I think this is unexpected hidden behavior that may trip someone using this script from a different then project root directory.
 
Helper scripts are nice. I think they should strive to be as correct as possible. 

This is maybe relevant for all the scripts, but most of the others make some sense to be run with different working directories.

scripts/lint.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [[ -z "${FILES}" ]]; then
+  echo "No source files found."
+  exit 0
+fi
+
+echo "Running clang-tidy..."
+clang-tidy -p "${BUILD_DIR}" ${FILES}
```
michalhosna: `${FILES}` is not quoted, so it gets globbed and word-split. That may not result in the correct command, especially when IFS is not explicitly set. 

There are many valid solutions (e.g. `find` with `-exec`, `xargs` or splitting at null bytes). Not suggesting anything concrete given the RG comment above.  

I would sugest adding shellcheck lint to CI (and maybe a helper script).

michalhosna (changes requested):
I would much rather see a CI setup so that usage of all the tools in CI is clear and enforces things right from the get-go than long-ish markdowns. CI can be postponed to another PR. But I expect we will require every PR to contain tests for (almost) everything that's changing. By that same logic, CI setup should be here so we are not merging something that doesn't work. But I am happy to support small atomic PRs.

Sorry if the review is too harsh; it's meant to be factual. I hope it gets better once we agree on the basic rules.

src/relay.cpp
```diff
@@ -0,0 +1,11 @@
+#include "o_rly/relay.h"
+
+namespace o_rly {
```
afrind: do we want our top level namespace to be 'openmoq' and this project be opemoq::o_rly

docker/Dockerfile
mondain: Based solely on percentages of running servers; Among the servers running Linux, Ubuntu is the clear leader for web infrastructure, while Debian remains a popular "stable" alternative.

Ubuntu: ~31.8% to 33.9% of Linux servers. It is the most popular choice for cloud-based virtual machines (AWS, Azure, Google Cloud) due to its ease of use and long-term support (LTS).

Debian: ~16.3% to 16.8% of Linux servers. Known for its "rock-solid" stability, it is often the foundation for other distributions (including Ubuntu) and is heavily used in high-security environments.

CentOS / RHEL: ~8% to 9%. While the original CentOS has been discontinued in favor of CentOS Stream, its legacy still accounts for a significant portion of older web servers. Red Hat Enterprise Linux (RHEL) actually holds a massive 43% share of the "paid" enterprise server market, even if its share of total web servers seems smaller.

design/ARCHITECTURE.md
mondain: Fair enough; I did not include that doc in the prompt; that is my mistake, but easily rectified.

design/ROADMAP.md
mondain: Yes, high level generation based on supporting the last 3 drafts, but then again we have to start somewhere.

.gitignore
mondain: I get what you're saying, but its an apples and oranges comparison. nginx is not a cmake styled build, its basic makefile style, so the ignores for it are quite different from cmake as a build directory is normally used and the artifacts would be contained within that location. In this first PR, I did not specifically include testing.

README.md
mondain: Its meant as the "reference" implementation; yet the wording may not translate as intended. I'm open to fully dropping that wording.

.gitattributes
mondain: Here is why .gitattributes is a lifesaver in a multi-OS environment.

1. The "End-of-Line" Nightmare (LF vs. CRLF)
This is the most common cross-platform headache.

Windows uses Carriage Return + Line Feed (CRLF).

macOS/Linux uses just Line Feed (LF).

Without .gitattributes, a Windows developer might open a file, save it, and suddenly Git thinks every single line has changed because the line endings were converted.

How .gitattributes fixes it: You can force Git to handle this automatically so the code is always stored as LF in the repository but matches the developer's OS locally:

Code snippet
# Force all text files to be normalized to LF in the repo
* text=auto
*.sh text eol=lf
*.bat text eol=crlf
2. Preventing Binary Corruption
Git sometimes tries to be "helpful" by normalizing line endings in files it thinks are text. If it accidentally treats a PNG, PDF, or Word Doc as text, it can modify the internal bytes to "fix" line endings, which corrupts the file and makes it unopenable.

How .gitattributes fixes it: You can explicitly tell Git which files are binary and should never be touched:

Code snippet
*.png binary
*.jpg binary
*.pdf binary
3. Syntax Highlighting and Diffing
Different OS tools and Git GUIs (like Sourcetree or GitKraken) might not know how to "diff" (compare) specialized files. .gitattributes allows you to tell Git how to treat specific files during a comparison.

Example: You can tell Git to ignore certain files in "diffs" to keep pull requests clean, or identify generated files (like min.js) so they don't clutter up the code review.

scripts/lint.sh
mondain: rg or find doesn't matter to me, ripgrep is simply a newer and very performant tool.

.gitignore
mondain: VS Code is my tool of choice; most of y'all use macOS so XCode is assumed and Tim uses CLion. We can certainly adjust any other entries.

src/relay.cpp
mondain: Good question, I think openmoq namespace may be a good idea; I defer to the group...

mondain:
> I would much rather see a CI setup so that usage of all the tools in CI is clear and enforces things right from the get-go than long-ish markdowns. CI can be postponed to another PR. But I expect we will require every PR to contain tests for (almost) everything that's changing. By that same logic, CI setup should be here so we are not merging something that doesn't work. But I am happy to support small atomic PRs.
> 
> Sorry if the review is too harsh; it's meant to be factual. I hope it gets better once we agree on the basic rules.

This layout is for us as developers working on the code with the assistance of our IDE's etc; honestly, no one mentioned CI or I somehow completely missed it? Either way, its not an impossible task to add the required steps to enable a CI or Github Actions.

design/ROADMAP.md
michalhosna: Sorry, I did not made myself clear. 

I think the whole design/ROADMAP.md is out-of-scope and imho should not be in this PR at all.

docker/Dockerfile
michalhosna: Numbers about what's installed as the base OS should not be relevant here. We are talking about the Docker base image, which may be entirely different and have entirely unique tradeoffs that we aren't concerned about here (e.g. init system). The point of containers in the first place is decoupling the base OS from the runtime. 

As I linked, Debian is the more used base. And there's hardly any Ubuntu in the official Docker library images. 

Sidenote: Please provide sources and references for your numbers.

scripts/lint.sh
michalhosna: We are building an open-source application, and we hope for wide adoption. Any non-standard dependency in the tooling we bring should have a good reason to be added. 

What's the good reason to add it here? Do you experience performance problems with find? 
I think "newer" tool is not a good argument in itself. 

If there's a good-enough alternative in POSIX or at least GNU *utils, I would definitely call that a more standard alternative, available essentially everywhere, than anything that has to be sourced from somewhere else.  This script now may not be run from the Dockerfile you provided. It'll require a custom image or install step in CI, etc...

Using grep alternatives seems improper when we are looking at the `find` use case; if for nothing else, the readability and intent of the script will be clearer with find. 

Don't get me wrong, I like `rg` and use it daily, but in a shared OSS repo, the bar should be higher than that.

.gitignore
michalhosna: From the official Git doc at https://git-scm.com/docs/gitignore (emphasis mine) 

> Which file to place a pattern in depends on how the pattern is meant to be used.
> 
> - Patterns which should be version-controlled and distributed to other repositories via clone (i.e., files that all developers will want to ignore) should go into a .gitignore file.
> 
> - Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are **specific to one user’s workflow**) should go into the $GIT_DIR/info/exclude file.
> 
> - Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the **user’s editor of choice**) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

The point with nginx is, there's nothing regarding IDEs there. We are creating an OSS project; we do not know what people will use, and the current set of people is irrelevant in that matter. I do not expect 3 line long file. 

If people don't know how to set up their local Git, we may have a guide in something like `CONTRIBUTING.md`. For example, point them to the relevant Git documentation. 

I know this may seem pedantic, but anything we submit to the repository we may have to maintain and support. `.gitignore` is a relatively critical piece of infrastructure, as pointed out above. The best code is no code.

design/ARCHITECTURE.md
michalhosna: Sorry, I did not made myself clear.

I think the whole design/ARCHITECTURE.md is out-of-scope and imho should not be in this PR at all. It not being correct is just another argument for removal.

scripts/lint.sh
```diff
@@ -0,0 +1,19 @@
...
+
+FILES=$(rg --files -g '*.{h,hpp,cc,cpp,cxx}' include src tests tools)
+
+if [[ -z "${FILES}" ]]; then
+  echo "No source files found."
+  exit 0
+fi
+
+echo "Running clang-tidy..."
+clang-tidy -p "${BUILD_DIR}" "${FILES}"
```
michalhosna: The `"${FILES}"` fix is imho , incorrect; you do need the file splitting; the problem is it has to happen at the right places. 

Does this script work for you? 

```suggestion
run-clang-tidy -p "${BUILD_DIR}" 
```

> Runs clang-tidy over all files in a compilation database.

This should do the same thing by parsing `compile_commands.json` and handling all the edge cases. Do we need to tidy files not present in the compilation database?

.gitattributes
michalhosna: This all seems like a problem that should be configured by the local Git config, not something that is project-specific. 

I would expect we need to override non-obvious defaults here. 
The defaults for newlines are to follow the existing convention in the repository. 
And marking the source files as text is also default. 

Looking at other bigger projects, 
- nginx, doesn't have this file at all
- [postgres/.gitattributes](https://github.com/postgres/postgres/blob/master/.gitattributes) does have this file, but only overrides non-default, non-obvious preferences. 

What's the benefit that we should have this file with generic content when other established projects don't need it? 
Can you point me to a reference that recommends your approach? 

Again, this may seem pedantic, but why are we adding a maintenance burden and unecessary files?

michalhosna (changes requested):

michalhosna:
> This layout is for us as developers working on the code with the assistance of our IDE's etc; honestly, no one mentioned CI or I somehow completely missed it? Either way, its not an impossible task to add the required steps to enable a CI or Github Actions.

I don't think we have a great problem assignment at the moment. But with project scaffold/setup, I would personally 100% expect a CI. It also serves as great documentation on how the projects should be built and run, i.e., how the scripts you are proposing should be used.

docker/Dockerfile
mondain: Source: https://sqmagazine.co.uk/linux-statistics/#:~:text=Government%2Dfunded%20HPC%20labs%20reported,stability%20and%20long%2Dterm%20support.

mondain: I prefer and it makes sense to me to build for the target platform with that platform's base image, but the difference is inconsequential here, so would the update to using `bookworm` satisfy your request?

.gitignore
mondain: Is the update sufficient? I kept `.DS_Store` as macos likes to leave turds here and there

.gitattributes
mondain: I'm fine deleting it; the reason for the file is enforcement by git. Why I recommend it, experience.

docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM bookworm
```
michalhosna:
```suggestion
FROM debian:13
```

Just `bookworm` doesn't exists. 
Also `trixie` is the current stable. 
Given that the names are non obvious maybe going with numbers is better here.

michalhosna (changes requested):

src/relay.cpp
```diff
@@ -0,0 +1,11 @@
...
+
+Relay::Relay() = default;
+
+std::string Relay::Version() const {
+  return "0.1.0";
+}
+
+}  // namespace o_rly
```
afrind:
```suggestion
namespace openmoq::o_rly {

Relay::Relay() = default;

std::string Relay::Version() const {
  return "0.1.0";
}

}  // namespace openmoq::o_rly
```

src/relay.cpp
afrind: eh, let's to it.

michalhosna (approved):
I am happy to leave CI for later so we are not blocking PR. I do expect that CI setup may end up changing a bunch of this stuff, but that's fine.

PR #10, opened 2026-02-04
Draft: Configuration and hot-reloading.

michalhosna:
I'd consider this a draft; Claude is too verbose for my liking, so I tried to shorten it. Inconsistencies popped up. 

Feedback on the direction is welcomed. 

I am not sure if we need to support binary reloading on systems where uBPF steering is unsupported. I don't think so; a load balancer before the relay can take care of that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/10)

### DIFF
--- a/design/configuration.example.yaml
+++ b/design/configuration.example.yaml
@@ -1,0 +1,87 @@
+
+global:
+  worker_threads: 4               # I/O threads; 0 = auto (num CPUs)
+  pid_file: /var/run/o-rly.pid
+
+  log_level: info                  # trace, debug, info, warn, error
+  max_connections: 100000          # Per-process connection limit
+
+  slow_subscriber_queue_max: 1000  # Max queued objects before terminating subscription
+  delivery_timeout_ms: 500         # Default delivery timeout
+
+  cache:
+    memory_max_mb: 4096
+    eviction_policy: lru           # lru, lfu
+    disk_path: ""                  # Empty = disabled
+    disk_max_mb: 0
+
+  telemetry:
+    metrics_endpoint: "0.0.0.0:9090"
+    metrics_format: opentelemetry
+    qlog_enabled: true
+    qlog_dir: /var/log/o-rly/qlog
+    mlog_enabled: true
+
+listeners:
+- name: main
+  address: "0.0.0.0"
+  port: 4433
+  protocols:
+  - quic
+  - webtransport
+  tls:
+    cert_file: /etc/o-rly/tls/cert.pem
+    key_file: /etc/o-rly/tls/key.pem
+    # ... Aditional tls config as necessary
+  quic:
+    congestion_control: bbr
+    initial_rtt_ms: 100
+
+- name: wt-main
+  address: "0.0.0.0"
+  port: 443
+  protocols:
+  - webtransport
+  tls:
+    cert_file: /etc/o-rly/tls/cert.pem
+    key_file: /etc/o-rly/tls/key.pem
+
+services:
+- name: live-streaming
+  authority: "live.example.com"
+  path: "/"
+
+  auth_plugins:
+      #......
+  max_sessions: 50000
+
+  provider:
+    type: static_upstream
+    config:
+      upstream_uri: "moqt://origin.example.com:4433"
+
+  cache:
+    memory_max_mb: 2048          # Override global cache for this service
+    eviction_policy: lfu
+
+  congestion_control: bbr        # Per-service override
+
+  namespaces:
+  - prefix: ["live", "sports"]
+    max_max_cache_duration_ms: 200
+    slow_subscriber_queue_max: 500
+
+  - prefix: ["live", "sports"]
+    max_max_cache_duration_ms: 2000
+
+- name: catch-all
+  authority: "*"
+  path: "/"
+  auth_plugins:
+  - anonymous
+
+  provider:
+    type: local-only
+
+services_include:
+- /etc/o-rly/services.d/*.yaml
--- a/design/configuration.example.yaml
+++ b/design/configuration.example.yaml
@@ -1,87 +1,99 @@
-
 global:
-  worker_threads: 4               # I/O threads; 0 = auto (num CPUs)
-  pid_file: /var/run/o-rly.pid
+  worker_threads: 4 # I/O threads; 0 = auto (num CPUs)
+  log_level: info
 
-  log_level: info                  # trace, debug, info, warn, error
-  max_connections: 100000          # Per-process connection limit
-
-  slow_subscriber_queue_max: 1000  # Max queued objects before terminating subscription
-  delivery_timeout_ms: 500         # Default delivery timeout
-
-  cache:
-    memory_max_mb: 4096
-    eviction_policy: lru           # lru, lfu
-    disk_path: ""                  # Empty = disabled
-    disk_max_mb: 0
-
-  telemetry:
-    metrics_endpoint: "0.0.0.0:9090"
-    metrics_format: opentelemetry
-    qlog_enabled: true
-    qlog_dir: /var/log/o-rly/qlog
-    mlog_enabled: true
+  telemetry: ....
 
 listeners:
-- name: main
-  address: "0.0.0.0"
-  port: 4433
-  protocols:
-  - quic
-  - webtransport
-  tls:
-    cert_file: /etc/o-rly/tls/cert.pem
-    key_file: /etc/o-rly/tls/key.pem
-    # ... Aditional tls config as necessary
-  quic:
-    congestion_control: bbr
-    initial_rtt_ms: 100
+  - name: main
+    udp: # Exactly one of udp, tcp (maybe pipe, qmux?)
+      socket:
+        address: "::"
+        port: 433
+        ipv6_only: false
+      transport:
+        mvfst: # implementation specific config
+          protocols:
+            quic:
+            webtransport:
+          tunables:
+            initial_congestion_control: bbr
+            initial_rtt_ms: 100
 
-- name: wt-main
-  address: "0.0.0.0"
-  port: 443
-  protocols:
-  - webtransport
-  tls:
-    cert_file: /etc/o-rly/tls/cert.pem
-    key_file: /etc/o-rly/tls/key.pem
+    tls_credentials:
+      # TODO: Tls credential providers
+      # - Do we support more then one provider at a time?
 
-services:
-- name: live-streaming
-  authority: "live.example.com"
-  path: "/"
+  - name: tcp-fallback
+    tcp:
+      socket:
+        address: "::"
+        port: 443
+      transport:
+        proxygen: # I don't know who will implement webtransport/h2
+    tls_credantials: ~ #TODO, see above
 
-  auth_plugins:
-      #......
+service_defaults:
+  slow_subscriber_queue_max_bytes: 1000
+  auth: {} #...
+  cache: {} #...
   max_sessions: 50000
 
-  provider:
-    type: static_upstream
-    config:
-      upstream_uri: "moqt://origin.example.com:4433"
+services:
+  - name: live-streaming
+    authorities:
+      - "live.example.com"
+      - "e79.x.akamaiedge.net"
+      - "1234567890.rsc.cdn77.org"
+    path:
+      prefix: "/"
 
-  cache:
-    memory_max_mb: 2048          # Override global cache for this service
-    eviction_policy: lfu
+    # Tune batching e.g. Latency vs performance
+    batching:
+      min_batch_size: 100kib
+      max_batch_delay: 250ms
 
-  congestion_control: bbr        # Per-service override
+    # By defualt all listners, this can limit them
+    listeners: ["main"]
 
-  namespaces:
-  - prefix: ["live", "sports"]
-    max_max_cache_duration_ms: 200
-    slow_subscriber_queue_max: 500
+    auth: #....
 
-  - prefix: ["live", "sports"]
-    max_max_cache_duration_ms: 2000
+    provider:
+      type: static_upstream
+      config:
+        upstream_uri: "moqt://origin.example.com:4433"
 
-- name: catch-all
-  authority: "*"
-  path: "/"
-  auth_plugins:
-  - anonymous
+    cache: #....
 
-  provider:
-    type: local-only
+    # TODO: How to handle per service-overrides?
+    congestion_control: bbr
+
+    tracks:
+      # All tracks in given namespace prefix
+      - namespace:
+          prefix: ["live", "sports"]
+        max_max_cache_duration_ms: 200
+        slow_subscriber_queue_max: 500
+
+      # I do thing that we need fairly advanced matching for examples MSF, where the catalog track may need different params. Is there a simplier way?
+      - namespace:
+          prefix:
+            - exact: "live"
+            - regex: "stream-[a-zA-Z0-9]+"
+        track:
+          exact: "catalog"
+        batching:
+          max_batch_delay: 50ms
+
+  - name: pure-relay
+    authority:
+      wildcard: "moq-relay.example.com"
+    path: "/"
+    auth_plugins:
+      - anonymous
+
+    provider:
+      type: local-only
 
 services_include:
-- /etc/o-rly/services.d/*.yaml
+  - /etc/o-rly/services.d/*.yaml
--- a/design/configuration.md
+++ b/design/configuration.md
@@ -1,0 +1,180 @@
+# Configuration & Hot-Reloading Design
+
+## Overview
+
+The relay needs both file-based configuration for base deployment and a runtime API for dynamic updates. Configuration must be hierarchical, mapping naturally to MOQT's connection model (authority, path, namespace).
+
+### Requirements References
+
+- [Relay Requirements 2.2](https://github.com/OpenMOQ/relay-requirements/blob/main/relay-requirements.md): "configurable via environment variables and/or a configuration file (e.g., YAML, TOML). Configuration should be reloadable without requiring a server restart where possible."
+- [Relay Requirements 3.3.2](https://github.com/OpenMOQ/relay-requirements/blob/main/relay-requirements.md): Control Plane API for management and orchestration.
+- [Relay Requirements 3.6](https://github.com/OpenMOQ/relay-requirements/blob/main/relay-requirements.md): Congestion control configurable at minimum per service.
+
+### Goals
+
+1. **File-based configuration** as the primary interface for base deployment.
+2. **Runtime API** for dynamic updates in orchestrated/CDN environments.
+3. **Hierarchical structure** that maps to MOQT connection model.
+4. **Clear reloadability semantics**: which parameters are static, which are hot-reloadable, and how reload affects existing connections/subscriptions.
+5. **Graceful binary restart** for deployments requiring process replacement.
+
+## Prior Art
+
+| Project | Config Format | Hierarchy | Dynamic Reload | Runtime API | Binary Restart |
+|---------|--------------|-----------|----------------|-------------|----------------|
+| **nginx** | Custom DSL | `http` → `server` → `location` | `SIGHUP`: new workers, old drain | nginx+ only (commercial) | New workers inherit listen sockets; old workers drain |
+| **Envoy** | YAML/JSON | `static` (bootstrap) + `dynamic` (xDS) | xDS API: listeners, clusters, routes, endpoints | Full xDS; also admin API | Hot restart via shared memory between old/new |
+| **HAProxy** | Custom DSL | `global` → `defaults` → `frontend` → `backend` | `-sf` soft reload; new process inherits sockets | Runtime API for limited changes (server weights, health) | Socket inheritance via Unix domain sockets |
+| **Caddy** | JSON + Caddyfile | Sites → routes → handlers | Full config reload via API, zero-downtime | First-class JSON API for all config | Socket inheritance via `SO_REUSEPORT` |
+| **Traefik** | YAML/TOML + providers | `static` (entrypoints, global) + `dynamic` (routers, services) | File provider watches for changes | Kubernetes/Docker/file providers | Relies on orchestrator (no built-in) |
+
+### Key Takeaways
+
+- **Envoy's static/dynamic split** is the strongest model for our use case: some things (listeners, thread count) are inherently static; routes, upstreams, and policies are inherently dynamic.
+- **Custom DSLs** (nginx, HAProxy) are powerful but create a learning barrier and maintenance burden. Both projects have paid the cost of building and maintaining parsers, and users frequently cite config syntax as a pain point.
+- **All mature proxies** distinguish between "needs new process" and "can update in-place" parameters.
+
+## Configuration Format
+
+### Decision: YAML
+
+**Chosen**: YAML as the file-based configuration format.
+
+**Rationale:**
+- Widely used in infrastructure tooling. Operators already know it.
+- Supported by mature C++ parsing libraries ([yaml-cpp](https://github.com/jbeder/yaml-cpp), [rapidyaml](https://github.com/biojppm/rapidyaml)).
+
+**YAML pitfalls** (e.g., the "Norway problem" where `NO` is parsed as boolean): mitigated by always deserializing into strongly-typed C++ structs. The relay never interprets raw YAML scalars as implicit types. Values are always parsed as the expected type defined in the config schema, or rejected. This is consistent with how Envoy and Kubernetes handle YAML.
+
+For the API we can use the same format, require JSON serialization or go with something completly else (e.g. JSON Patch for updating). 
+
+## Configuration Hierarchy
+
+The hierarchy should map to the MOQT connection model.
+
+```
+global          ─── process-wide settings
+├── listeners   ─── listen address/port/protocol/TLS (maps to transport layer)
+└── services    ─── matched by MOQT authority + path (maps to CLIENT_SETUP)
+    └── namespaces ─── matched by track namespace prefix (maps to SUBSCRIBE/PUBLISH)
+```
+
+### Matching Model
+
+A client connects to a **listener** (transport), then sends `CLIENT_SETUP` with **authority** and **path**. These are matched against **services**. Subsequent `SUBSCRIBE`/`PUBLISH` messages carry track namespaces, matched against **namespace** rules within the matched service.
+
+**Matching priority** (highest to lowest):
+1. **Exact match**: `authority: "live.example.com"`
+2. **Wildcard prefix**: `authority: "*.example.com"`
+3. **Regex**: `authority_regex: "^[a-z]+-region[0-9]+\\.example\\.com$"`
+4. **Default/catch-all**: `authority: "*"`
+
+Same priority order applies to `path` and `namespace_prefix`.
+
+> **Performance note**: Regex matching is evaluated only when exact and prefix matches fail. For high-throughput paths (per-object routing), prefer prefix matching. Regex is appropriate for service-level matching (per-connection), which happens once at session setup. Namespace-level regex should be used sparingly and is explicitly not evaluated per-object — only at subscription setup time.
+
+> **MOQT Scope note**: The combination of authority + path defines a MOQT scope ([moq-transport#1432](https://github.com/moq-wg/moq-transport/issues/1432). Sessions within the same scope share namespace trees and cache. The config hierarchy reflects this: each service (authority + path pair) is an isolated scope by default. Cross-scope sharing (e.g., shared cache across authorities) can be configured explicitly via named shared resources.
+
+### Inheritance
+
+Configuration parameters are inherited down the hierarchy. A more specific level overrides its parent. This follows the nginx model where `server` inherits from `http`, and `location` overrides `server`.
+
+### Include Directives
+
+Includes are supported **only at the service level**: a `services_include` directive loads service definitions from external files. This keeps the main config file manageable while preventing fragmentation of the hierarchy at arbitrary levels.
+
+```yaml
+services_include:
+  - /etc/o-rly/services.d/*.yaml  # Each file defines one or more services
+```
+
+Each included file must contain complete service definitions. Partial overrides or nested includes are not supported. This is intentional.
+
+## Configuration Schema
+
+### Parameter Lifecycle Classification
+
+Every configuration parameter belongs to one of four lifecycle categories. The config schema documentation (generated from the C++ struct definitions) must annotate each parameter with its lifecycle.
+
+| Lifecycle | Meaning | Examples                                                                            |
+|-----------|---------|-------------------------------------------------------------------------------------|
+| **Static** | Requires process restart. Cannot be changed at runtime. | `worker_threads`, `listeners[].address`, `listeners[].port`|
+| **Reload:NewConn** | Applied on config reload. Affects new connections/sessions only. Existing sessions keep old values. | `services[].auth_plugin`, `services[].provider`, TLS cert/key, service matching rules |
+| **Reload:NewSub** | Applied on config reload. Affects new subscriptions/tracks only. Existing subscriptions keep old values. | `namespaces[].max_max_cache_duration_ms`, routing rules, authorization rules        |
+| **Reload:Immediate** | Applied on config reload. Takes effect on all existing connections/subscriptions. | `log_level`, `slow_subscriber_queue_max`, cache eviction policy, metrics config     |
+
+## Design Decisions
+
+### YAML Over Custom DSL
+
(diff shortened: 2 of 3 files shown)


design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
+
+global:
+  worker_threads: 4               # I/O threads; 0 = auto (num CPUs)
+  pid_file: /var/run/o-rly.pid
+
+  log_level: info                  # trace, debug, info, warn, error
```
afrind: we can support > 1 debug level eg DBG0...N

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+  max_connections: 100000          # Per-process connection limit
```
afrind: inbound, outbound or both?  is it a file descriptor limit or QUIC/HTTP connections?

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+  slow_subscriber_queue_max: 1000  # Max queued objects before terminating subscription
```
afrind: I was confused at first but see this is "default if not overriden".

We could also express a total amount of object memory per service which is spread over the cache and queues -- though it's harder to track.

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+  delivery_timeout_ms: 500         # Default delivery timeout
```
afrind: I don't think per MOQT we can specify a default - the publishers and subscribers set this and have their own expectations.  Could be useful is when o_rly is acting as a subscriber? I think we need to upstream timeout to be at least the max of the downstream timeouts - or there can be a cap.

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+  cache:
```
afrind: Let's call it memory_cache, because disk cache is a plug-in, right?

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+    memory_max_mb: 4096
```
afrind: Is this global or per-service?

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+    eviction_policy: lru           # lru, lfu
```
afrind: In MOQT terms you may want to have separate eviction policies for tracks and groups within a track.  eg: cache at most 1000 tracks, 3 groups / track.

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+    disk_path: ""                  # Empty = disabled
+    disk_max_mb: 0
+
+  telemetry:
+    metrics_endpoint: "0.0.0.0:9090"
+    metrics_format: opentelemetry
+    qlog_enabled: true
```
afrind: Does this generate a qlog for everything?

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+    qlog_dir: /var/log/o-rly/qlog
+    mlog_enabled: true
+
+listeners:
```
afrind, line 7: What's the relationship between listeners and services?  All services are accepted on all listeners, or a subset?  In configs where a service has a dedicated listener it might feel more natural for listeners to be nested under service.

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+- name: main
+  address: "0.0.0.0"
```
afrind: How do we want to support multiple addresses per listener (eg v4, v6, zero rated, etc)

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+  port: 4433
+  protocols:
+  - quic
+  - webtransport
```
afrind: webtransport/h3
webtransport/h2
qmux

Do we need quic and/or WT versions also?

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+  tls:
+    cert_file: /etc/o-rly/tls/cert.pem
+    key_file: /etc/o-rly/tls/key.pem
+    # ... Aditional tls config as necessary
+  quic:
+    congestion_control: bbr
```
afrind: quic_impl = mvfst | pico | ...

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+    initial_rtt_ms: 100
```
afrind: Plus other quic settings (flow control limits, TP, other tunables)

design/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+| **Envoy** | YAML/JSON | `static` (bootstrap) + `dynamic` (xDS) | xDS API: listeners, clusters, routes, endpoints | Full xDS; also admin API | Hot restart via shared memory between old/new |
+| **HAProxy** | Custom DSL | `global` → `defaults` → `frontend` → `backend` | `-sf` soft reload; new process inherits sockets | Runtime API for limited changes (server weights, health) | Socket inheritance via Unix domain sockets |
+| **Caddy** | JSON + Caddyfile | Sites → routes → handlers | Full config reload via API, zero-downtime | First-class JSON API for all config | Socket inheritance via `SO_REUSEPORT` |
+| **Traefik** | YAML/TOML + providers | `static` (entrypoints, global) + `dynamic` (routers, services) | File provider watches for changes | Kubernetes/Docker/file providers | Relies on orchestrator (no built-in) |
+
+### Key Takeaways
+
+- **Envoy's static/dynamic split** is the strongest model for our use case: some things (listeners, thread count) are inherently static; routes, upstreams, and policies are inherently dynamic.
```
afrind, line 39: Why are listeners inherently static?  When a new customer signs up are they not adding a listener?  Or perhaps the listen part is static, but e.g. the certificate / key database is somehow dynamic when it receives unknown SNI?

design/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+### Matching Model
+
+A client connects to a **listener** (transport), then sends `CLIENT_SETUP` with **authority** and **path**. These are matched against **services**. Subsequent `SUBSCRIBE`/`PUBLISH` messages carry track namespaces, matched against **namespace** rules within the matched service.
+
+**Matching priority** (highest to lowest):
+1. **Exact match**: `authority: "live.example.com"`
+2. **Wildcard prefix**: `authority: "*.example.com"`
+3. **Regex**: `authority_regex: "^[a-z]+-region[0-9]+\\.example\\.com$"`
```
afrind: Is there a universe where we don't have to support full regex matching and can get away with starts-with/ends-with?

Do we need composite rules (live.example.com && (path=/foo || path=/bar)) ?

design/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+4. **Default/catch-all**: `authority: "*"`
+
+Same priority order applies to `path` and `namespace_prefix`.
```
afrind: Do we want special namespace handling match logic to factor in tuple boundaries?

design/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+
+> **Performance note**: Regex matching is evaluated only when exact and prefix matches fail. For high-throughput paths (per-object routing), prefer prefix matching. Regex is appropriate for service-level matching (per-connection), which happens once at session setup. Namespace-level regex should be used sparingly and is explicitly not evaluated per-object — only at subscription setup time.
```
afrind: I don't think we need "per object routing" - I hope!  MOQT data plane is established by control messages and objects just follow the established plane.

design/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+
+> **MOQT Scope note**: The combination of authority + path defines a MOQT scope ([moq-transport#1432](https://github.com/moq-wg/moq-transport/issues/1432). Sessions within the same scope share namespace trees and cache. The config hierarchy reflects this: each service (authority + path pair) is an isolated scope by default. Cross-scope sharing (e.g., shared cache across authorities) can be configured explicitly via named shared resources.
```
afrind: What's an example of how we will support named shared resources?

design/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+
+Configuration parameters are inherited down the hierarchy. A more specific level overrides its parent. This follows the nginx model where `server` inherits from `http`, and `location` overrides `server`.
+
+### Include Directives
+
+Includes are supported **only at the service level**: a `services_include` directive loads service definitions from external files. This keeps the main config file manageable while preventing fragmentation of the hierarchy at arbitrary levels.
+
+```yaml
+services_include:
+  - /etc/o-rly/services.d/*.yaml  # Each file defines one or more services
```
afrind: Do we monitor this dir for new files and add them automatically?

design/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+Each included file must contain complete service definitions. Partial overrides or nested includes are not supported. This is intentional.
+
+## Configuration Schema
+
+### Parameter Lifecycle Classification
+
+Every configuration parameter belongs to one of four lifecycle categories. The config schema documentation (generated from the C++ struct definitions) must annotate each parameter with its lifecycle.
```
afrind, line 116: Love this

design/configuration.md
afrind: Ah, I see -- explicit signal

design/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+Regex matching is available at all hierarchy levels but is documented as a last resort:
+
+- **Service matching** (per-connection): Regex is acceptable. Evaluated once at session setup.
+- **Namespace matching** (per-subscription): Regex is acceptable but should be simple. Evaluated once at subscription setup.
+
+Matching order within a hierarchy level: exact → prefix (longest match) → regex (first match, config order) → default. This is the same precedence as nginx's location matching.
+
+### Config File Persistence from API
+
+**Chosen**: API changes persist to the config file by default.
```
afrind, line 169: Ok but you can disable this behavior?

design/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+
+**Alternative**: API is ephemeral (HAProxy model).
+
+Persistence prevents config drift, which is the #1 operational problem with HAProxy's runtime API. If an operator modifies a parameter via the API and the process restarts, the change is preserved. This can be disabled for ephemeral/immutable-infrastructure deployments where config is injected at startup and never modified in-place.
+
+When persistence is enabled, the config file is rewritten atomically (write to temp file, then `rename(2)`) after each API mutation.
```
afrind, line 175: Does it save the old ones?

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+  slow_subscriber_queue_max: 1000  # Max queued objects before terminating subscription
+  delivery_timeout_ms: 500         # Default delivery timeout
+
+  cache:
+    memory_max_mb: 4096
+    eviction_policy: lru           # lru, lfu
+    disk_path: ""                  # Empty = disabled
+    disk_max_mb: 0
+
+  telemetry:
```
afrind: We also need access log rules - location, frequency, metrics to include.

design/configuration.example.yaml
```diff
@@ -0,0 +1,87 @@
...
+  max_sessions: 50000
+
+  provider:
+    type: static_upstream
+    config:
+      upstream_uri: "moqt://origin.example.com:4433"
+
+  cache:
+    memory_max_mb: 2048          # Override global cache for this service
+    eviction_policy: lfu
```
afrind: I was thinking you could configure like:

```
cache:
  chain: memory memcache <custom plugin> disk

cache_memory:
  # memory cache settings

cache_disk:
  # Disk cache settings

cache_memcache:
  # memcache settings

cache_plugin1:
  # etc
```

afrind (commented):
Thanks for doing this - really great start.

michalhosna:
@afrind Thanks for the thorough review! I want to check what's the best way forward. I see two types of issues here. 

1. Issues regarding this PR topic directly, e.g. https://github.com/openmoq/o-rly/pull/10#discussion_r2771499593)
2. Issues with specific config values, e.g. log levels and log config. 

My goal here wasn't to flesh out every possible config value. But have a generic discussion about how to structure the config (hierarchy and matching), hot reloading, and lifecycles.  My bad, I should have made the config example much shorter and cleaner. 

Do we want to split the 2nd kind of issues elsewhere, e.g., logging configuration, or do we want to flesh out the whole config, other than, like, most of the config we can think of now here? 

Guidance wanted. 

My inclination is to go with splitting out specifics elsewhere. Mainly because I don't think GitHub is handling bigger reviews super great. 

But fleshing out everything here also makes sense, but then maybe I should directly do and implement some kind of parser/API/validator. Because how do we even want to document the config? I don't know about great solutions  for that either. 

-----

EDIT: Decided to remove the bad examples from the configuration.yaml, so we can focus on the configuration things.

design/configuration.example.yaml
michalhosna: Given that all of this is stack specific, should this be stack scoped/stack nested? 

Generally I don't have a favorite way in YAML to express this kind of implementation choice + implementation config "sum type". 

What's the clearest?

```
quic: 
    quic_impl: mvfst
    mvfst_config:
        congestion_control: bbr
````

or

````
quic:
    mvfst:
        congestion_control: bbr
```
    
or single implementation only.

````
quic: 
    quic_impl: mvfst
    config:
        congestion_control: bbr
```

design/configuration.example.yaml
michalhosna: Do we want to have runtime choice per listener, or is QUIC stack a compile-time pick-one choice? 

I always thought about pick-one at compile time. But compiling-in multiple with runtime/boot-time choice is probably more sensible?

design/configuration.md
michalhosna: You could probably fairly easily add new listeners dynamically. But changing existing ones is problematic. E.g., changing the QUIC stack, changing the listener address, etc. It's essentially a full restart anyway. 

TLS must definitely be dynamic and pluggable. I wasn't sure where/how to place the basic TLS config currently; it's the config of the listener, not the service. But it's a dynamic part of the listener. 

Generally, the lifetimes/dynamicity is IMHO the hard problem here. I would love to have a way in the config to be more obvious about which is which, but structuring the config around config lifetimes is very different from this current hierarchical/logic approach.

design/configuration.md
michalhosna: I would think having some kind of standard regex support (note [ReDoS](https://en.wikipedia.org/wiki/ReDoS)) would be the best here. 

It offers a lot of flexibility, and I would hope even authority+path matching can be done performantly enough to not be a problem. 

I would even expect maybe a capture group and variable substitution as a later addition. 

Not blocking on this; starting with exact matches and prefixes is fine. 

Do you have something specific against regex? 😄 
(This is more for a spec discussion.)  

----

> Do we need composite rules (live.example.com && (path=/foo || path=/bar))?

I was considering writing that down like so:

```yaml
name: live.example.com
authority: "live.example.com"
path: "^/(foo|bar)$"
```

Well, "need", I think it makes sense to support them, but I don't have a need at the moment.

design/configuration.md
michalhosna: Yes. 
The following few lines talks about Disabling this and "When persistance is enabled"

design/configuration.md
michalhosna: We may advise more features, but I feel like only updating the current one is fine.
The whole point here is not to have a difference between on-disk config and in-memory config. So that debugging is easier and restarts behave as expected. I don't think keeping history is in scope. It can be easily done with external systems (e.g. git). 

Do you have something specific in mind?

design/configuration.md
michalhosna: You may want to have multiple authorities or multiple authority + path combinations to share scope. 

It mainly solves the problem, that authority, i.e., domain, may be used for other things in DNS (e.g. different balancing) than only for content identification. 

We see this sometimes in the HTTP world with our customers; I would imagine here to be the same.

I think the resolution of https://github.com/moq-wg/moq-transport/issues/1432 is needed for full resolution here.

design/configuration.md
michalhosna: Fixed?

design/configuration.md
michalhosna: Like, more specific (i.e., longer tuples) prefixes should win? Yes, I do think so. 

Some for domains. Fixed

michalhosna:
Self note: The config needs a clear distinction between regex, wildcard, and exact match strings.

michalhosna:
Per-service override of tunables
The problem: We want overrides of transport params/tunable per service. E.g. congestion controller. 
One service maybe listnening on multiple listners. Minimally we need to suport TCP fallback for the same service with shared cache and scope.

We may want to set different tunable values for each service + listner combinations. The set of tunables may be different. Even the set is non-obvious as each quic implementation may have their own set of tunables and values. Not all of these tunables can be changed per service (i.e. after QUIC establishmend and CLIENT_SETUP)

What's the best way to do this?

design/configuration.example.yaml
michalhosna: Calling this out-of-scope, I think this should be covered by other PRs. I removed it.

design/configuration.example.yaml
michalhosna: I added a config parameter to services so they can define on which listeners they are valid. 

I expect many more services compared to listeners, so putting it there makes sense. 
By default I think service should be available to all listeners.

design/configuration.example.yaml
michalhosna: Calling this out-of-scope, I think this should be covered by other PRs. I removed it.

design/configuration.example.yaml
michalhosna: Calling this out-of-scope, I think this should be covered by other PRs. I removed it.

design/configuration.example.yaml
michalhosna: Calling this out-of-scope, I think this should be covered by other PRs. I removed it.

design/configuration.example.yaml
michalhosna: I think single listner should define a single "socket", so only one address is supported. 

I added ipv6 with ipv6_only flag example.

design/configuration.example.yaml
michalhosna: Add implementation selection with tunables.

design/configuration.example.yaml
michalhosna: I am calling this out-of-scope (to limit scope of this PR), I think this should be covered by other PRs. I removed it.

design/configuration.example.yaml
```diff
@@ -0,0 +1,105 @@
+global:
+  worker_threads: 4 # I/O threads; 0 = auto (num CPUs)
+  log_level: info
+
+  telemetry: ....
+
+listeners:
+  - name: main
+    udp: # Exactly one of udp, tcp (maybe pipe, qmux?)
```
afrind, line 9: It's not really udp/tcp but QUIC and TCP+TLS (pipe seems fine)?  See also the discussion related to URLs - then QUIC listener can list supported alpns (moqt/1, moqt/2, h3) and TCP+TLS would have h2 (qmux?).

I guess if a service does not want webtrans or native quic support but it gets negotiated due to limited client ALPNs, we can fail slightly later?

design/configuration.example.yaml
```diff
@@ -0,0 +1,105 @@
...
+      transport:
+        mvfst: # implementation specific config
+          protocols:
+            quic:
+            webtransport:
+          tunables:
+            initial_congestion_control: bbr
+            initial_rtt_ms: 100
+
+    # We could have different MOQT implementation per SNI or different between WT and QUIC. I am not sure that's really useful?
```
afrind: At this point moqt is not designed to be pluggable, but I'm not against having a config for it if we anticipate it later?

design/configuration.example.yaml
```diff
@@ -0,0 +1,105 @@
...
+    moqt_implementation:
+      moxygen: ... # Moxygen tunables required before client intilializaition
+    tls_credantials:
```
afrind:
```suggestion
    tls_credentials:
```

design/configuration.example.yaml
```diff
@@ -0,0 +1,105 @@
...
+  - name: tcp-fallback
+    tcp:
+      socket:
+        address: "::"
+        port: 443
+      transport:
+        proxygen: # I don't know who will implement webtransport/h2
+    moqt_implementation:
+      libquicr: ... # Moxygen tunables
+    tls_credantials: ~ #TODO, see above
```
afrind, line 34:
```suggestion
    tls_credentials: ~ #TODO, see above
```

design/configuration.example.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - "1234567890.rsc.cdn77.org"
+    path:
+      prefix: "/"
+
+    # Tune batching e.g. Latency vs performance
+    batching:
+      min_batch_size: 100kib 
+      max_batch_delay: 250ms
+
+    # By defualt all listners, this can limit them
```
afrind, line 56:
```suggestion
    # By default all listeners, this can limit them
```

design/configuration.md
```diff
@@ -0,0 +1,202 @@
+# Configuration & Hot-Reloading Design
+
+## TODO
+
+- Single MOQT implementation per listner?
+  - Different ALPNs and SNI could have different moqt handlers.
+- In "panel-to-audience" model (i.e. few real-time participatns, many low-latency viewers), how to define different tunables for the panel as oposite to the audience?
+
+We could do this just
```
afrind: ?

design/configuration.md
```diff
@@ -0,0 +1,202 @@
+# Configuration & Hot-Reloading Design
+
+## TODO
+
+- Single MOQT implementation per listner?
+  - Different ALPNs and SNI could have different moqt handlers.
+- In "panel-to-audience" model (i.e. few real-time participatns, many low-latency viewers), how to define different tunables for the panel as oposite to the audience?
```
afrind, line 7: Is the way to handle this by allowing the client applications to select server tunables via negotiated params?  Which tunables in particular are you expecting to change?

design/configuration.md
afrind: I'm not sure I'm totally following your last comment.  We probably need more thinking about how to do the dynamic certificate loading pieces since this is likely to need to plug-in as well.

design/configuration.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+### Matching Model
+
+A client connects to a **listener** (transport), then sends `CLIENT_SETUP` with **authority** and **path**. These are matched against **services**.
+
+**Matching priority** (highest to lowest):
+
+1. **Exact match**: `authorities: ["live.example.com"]`
+2. **Wildcard**: `authority: { wildcard: "moq-relay.example.com" }`
```
afrind: what's the wildcard here?

design/configuration.md
afrind: Just leaving a trail of breadcrumbs that can be used with logs to figure out what the config was at a given time for debug.

design/hot-reloading.md
```diff
@@ -0,0 +1,43 @@
...
+
+Sending `SIGHUP` to the relay process triggers config reload, it does never trigger graceful restart.
+Attempting to change static config triggers an error. 
+
+### 2. Runtime API
+
+The relay exposes an HTTP API (per [Relay Requirements 3.3.2](https://github.com/OpenMOQ/relay-requirements/blob/main/relay-requirements.md)) for programmatic configuration. This API is served on a separate listener (the management/control plane listener) and is authenticated via mTLS or API key.
+
+```
+GET    /v1/config                    # Return current running config (JSON)
```
afrind, line 20: DO these return yaml or json or .. ?

design/hot-reloading.md
```diff
@@ -0,0 +1,43 @@
...
+
+**eBPF-based steering**: An eBPF program attached to the UDP socket steers packets for existing connection IDs to the old process and new connections to the new process.
+
+> **Note**: We will likely need eBPF-based steering between I/O threads anyway for QUIC connection affinity (ensuring all packets for a given connection are processed by the same thread). Extending this to process-level steering during restarts is incremental work on top of that foundation. See [QUIC-LB](https://datatracker.ietf.org/doc/draft-ietf-quic-load-balancers/) and [Relay Requirements §3.7](https://github.com/OpenMOQ/relay-requirements/blob/main/relay-requirements.md) for the connection ID generation API requirement.
+
+**State handoff**: We explicitly do **not** attempt to transfer in-memory state (cache, subscription state) between old and new processes. This is complex, error-prone, and the correct approach for stateless horizontal scaling is to rely on upstream re-subscription and cache warming. GOAWAY gives clients the signal to reconnect to the new process.
```
afrind, line 43: The only one we may need to consider is in-memory cache, but perhaps that can be delegated to a shared-memory caching provider since it's inherently more complex.

design/hot-reloading.md
michalhosna: JSON. 

YAML is designed as JSON superset. 
So returning JSON for better machine support but accepting YAML for better readability (and comments!) makes sense in all places.

design/configuration.md
michalhosna: Would logging a hash of the file be sufficient for that?
Or logging the whole file content? Logging a diff? 

Saving a new file each time seems not really a great solution to me. You have to manage cleanup, disk storage, etc. 

Delegating this to a logging system seems more reasonable.

design/configuration.md
michalhosna: The listeners are static; the TLS part needs to be dynamic. 

For TLS specifically, I think we should design a plugin system for it, but I'd consider that out of scope for this PR.

The hard problem I was referring to is the lifecycle of config parameters: different values apply at different times, some only take effect for new connections, some for existing connections, some for existing subscribers, and some only after a full application restart. 

Ideally, it would be obvious from the config file itself which lifecycle applies to which value. The only way I can think of, to achieve that, is a top-level structural split, section for each lifecycle. 

The current hierarchical structure is a good mental model for the logical organization of config, but it doesn't capture lifecycle at all, which means that has to live in documentation elsewhere rather than being self-evident from the schema.

Is it clearer, or am I causing more confusion?

design/configuration.md
michalhosna: I think it's for a spec discussion if and how we need to support this, if at all. 

I would expect we end up with a tunable for performance vs. latency, for example, batching objects/packets.

design/configuration.example.yaml
michalhosna: I am not sure if I am following your comment. 

I was thinking about UDP/TCP because of the split. What does the relay do, and what does the transport implementation do?

As the relay needs to set up a UDP socket with given params and then hand off the handling to a QUIC implementation. The ALPN is then handled by that QUIC+TLS implementation. Given that the QUIC implementation is something outside of the relay codebase, it doesn't have to be QUIC at all. 

A single listener should represent a single socket, i.e., single data entry point into the application. 

But I don't really care if we name it UDP or QUIC.

design/configuration.md
michalhosna: Git miscommit. I am genuinely sorry; I am fighting a little bit with the GitHub UI as it's hiding the comments I want to get back to.

design/hot-reloading.md
michalhosna: +1 on delegating to a shared-memory caching provider. 

Any handoff or sharing between binaries during restart/upgrade requires API/ABI compatibility, limiting the usefulness and coming with performance tradeoffs.

design/hot-reloading.md
```diff
@@ -0,0 +1,43 @@
+
+
+## Hot-Reloading Design
+
+### Terminology
+
+- **Config reload**: Re-reading and applying configuration without stopping the process. May affect only new or all connections depending on parameter lifecycle.
```
TimEvens, line 7: Config syntax should be commented and clear on which are run-time reloadable vs restart required.  For example, see postgres config.

design/hot-reloading.md
```diff
@@ -0,0 +1,43 @@
...
+- **Graceful restart**: Replacing the process binary. Old process drains existing connections; new process accepts new ones.
+
+### 1. File-Based Config Reload (SIGHUP)
+
+Sending `SIGHUP` to the relay process triggers config reload, it does never trigger graceful restart.
```
TimEvens:
```suggestion
Sending `SIGHUP` to the relay process triggers config reload only.
```

Often for drain and graceful restart, SIGUSR1 or similar would be used.

design/hot-reloading.md
TimEvens: Conversion from YAML is 1:1 to JSON.   It's not the other way around, such as YAML anchors/references, comments, ... IMHO, from an API perspective, JSON is more common and preferred. Config syntax documentation, file configs, etc. are often YAML formatted.

TimEvens (approved):
I think this is a good start.  I added some comments... Per discussion, we can iterate on this to fine tuning this.

design/configuration.example.yaml
afrind: I think I prefer quic, but not going to block on it now.

design/configuration.md
afrind: It's fine to document the behavior.  No need to reorganize the file.  +1 on the dynamic TLS bits which we discussed in the call.

design/configuration.md
afrind: Logging is fine as long as we also log the contents or patch itself somewhere.  e.g. if the relay crashed and the config has changed since, I need to know what config the relay was running with then it crashed _somehow_.

design/hot-reloading.md
afrind: I think there's a tradeoff between structuring the config hierarchically, which maps to the mental model, vs structuring it by reloadability and prefer the former.

afrind (approved):

design/hot-reloading.md
michalhosna: @TimEvens So do you want a wording change here, or are you okay with the current state?

design/hot-reloading.md
michalhosna: The draining is, IMHO, more complicated. I spun that off to a separate issue: https://github.com/openmoq/o-rly/issues/15

michalhosna (commented):
> @michalhosna made 2 comments, resolved 17 discussions, and dismissed @afrind from 17 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/10)* status: 0 of 3 files reviewed, 19 unresolved discussions (waiting on afrind and TimEvens).

michalhosna:
Should we merge this and iterate on implementation, or does anyone want changes here before merging and iterating on implementation?

cc @TimEvens @afrind

afrind:
Left some detailed comments in reviewable.  I think another pass might be good?  Feel free to ping with questions.

michalhosna:
I do not see any new comments. 
Are you talking about the old comments, or did you not publish them yet?

design/configuration.md
afrind: LGTM
Can follow up in spec

design/hot-reloading.md
afrind: LGTM

design/configuration.md
afrind: LGTM

design/configuration.md
afrind: LGTM

design/configuration.md
afrind: This wasn't quite what I meant, though I may have been confused.  Here you are only processing authority and path, so tuple boundaries are not part of it.  I meant for anything that matches namespaces, we need a mechanism to convey where the tuple boundaries are in the config specified match strings.

design/configuration.md
afrind: I guess so?

design/configuration.example.yaml
afrind: As much as we try there will be some params that are only supported for one stack.  I think I like a combo of 2 and 3 - eg: for things that every QUIC stack is going to support (flow control, congestion control named algos) - there's a single section.  Then there's also an impl specific section too (which is hopefully short).

afrind (approved):
I resolved all my outstanding comments.  I have a list of follow ups:

Regex Requirements
Named Shared Resources
Single/Multi Socket listeners
Transport specific tunables

> @afrind reviewed 3 files and all commit messages, made 8 comments, and resolved 15 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/10#-:-OmtaOfpDoMHst8p9n0u:b-w7gik4)* status: all files reviewed, 4 unresolved discussions (waiting on michalhosna and TimEvens).

PR #11, opened 2026-02-18
Integrate moxygen into o-rly

afrind:
There's a series of commits for more compact review

1. Fixes to existing build (missing .h and namespace problems)
2. Add the submodule (points to a branch with other fixes for now)
3. Build scripts
4. Adding moxygen into o-rly
5. Docker build

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.scratch/
+build/
+build-san/
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,8 @@ DartConfiguration.tcl
 # CPack
 _CPack_Packages/
 
+# Dependency scratch dir
+/.scratch/
+
 # OS
 .DS_Store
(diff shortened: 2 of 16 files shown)


gmarzot (approved):
All the getdeps and deep build support scripts are for the case where a dev needs to do the full (both moxygen and o-rly) but we expect when the local fork is doing all that is intended then that kind of build will be rare and unnecessary for the most part right. There will be ci build s which leverage the upstream moxygen (local buffer fork) build artifacts and this can also work for local dev builds too

michalhosna (approved):
Not blocking: It would be nice to update README. Point out submodules and build scripts.

afrind:
> There will be ci build s which leverage the upstream moxygen (local buffer fork) build artifacts and this can also work for local dev builds too

yes, for the local build I would like CI to export a build artifact of the deps just like docker cache so we don't all need to rebuild the world.

PR #12, opened 2026-02-18
Copy MoQRelay as ORelay and build a real relay server

afrind:
Hard-fork moxygen's MoQRelay into ORelay (openmoq::o_rly namespace) so we can evolve the relay core independently. MoQForwarder, MoQCache, MoQSession, and MoQServer continue to be used as libraries from moxygen.

- ORelay.h/cpp: copy+rename of MoQRelay with moxygen:: qualified types
- ORelayServer.h/cpp: extends MoQServer, wires ORelay as pub/sub handler
- main.cpp: real gflags-based relay main (port, cert, key, insecure, etc.)
- ORelayTest.cpp: full 33-test suite ported from MoQRelayTest
- Delete old stub relay.h/cpp and smoke_test
- CMakeLists.txt: link sub-deps, gtest_discover_tests for per-case enumeration


afrind:
There's an outstanding moxygen draft-16 change in the relay that we will need to port here when it lands.

include/o_rly/ORelay.h
mondain, line 1: Not a lawyer; I thought the licensing was to be Apache 2? or was it in-fact MIT as stated herein. Maybe the "Copyright xx Meta...." needs a clear blurb about what is actually copyrighted to Meta?

PR #28, opened 2026-03-02
Reformat

michalhosna:
EDIT: See #31 first

Result of running `scripts/format.sh`

I don't really like the result, but I also don't really care. We just should enforce some kind of formatting. 
We need CI enforcement of formatting soon, as this should have never get to main without proper formatting. 
I am not sure about the #13 state. Giovanni working hard on CI so I would rather not break his things. 

I think it's fine to merge this in its current form so that main has consistent `scripts/format.sh` and the actual files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/28)

PR #29, opened 2026-03-02
Implement basic config loading and parsing

### LINKED ISSUE #22: Basic configuration file and parsing
We need to have a config file, and parse it.

See proposed design in #10

michalhosna:
Based on #10. Replaces existing CLI flags with a config file. 
I decided to go with reflect-cpp as it is "self-documenting" with possible JSON schema generation. 
I'll follow up with implementing the rest of #10 after the basics are agreed upon here. 

Stands on top of #28 to use the automatic formatting. 
Fixes #22

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29)


mondain (approved):
> @mondain reviewed 9 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

afrind (commented):
> @afrind made 19 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29)* status: all files reviewed, 19 unresolved discussions (waiting on michalhosna).

___
*[`include/o_rly/config/config.h` line 13 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqGreI-KlAQI6_sQdG:-OmqGreI-KlAQI6_sQdH:b-lqyjh6) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L13)):*
> ```objc
> #include "o_rly/config/string_literal.h"
> 
> namespace openmoq::o_rly::config {
> ```

Do we like deeply nested namespaces? I 'm ok with it, just want to ask
___
*[`include/o_rly/config/config.h` line 19 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqH3Av7ukfrw9doWbq:-OmqH3Av7ukfrw9doWbr:b-754xx3) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L19)):*
> ```objc
>   static constexpr uint16_t kDefaultPort = 9668;
> 
>   rfl::Description<str_concat<"Bind address (default: \"", kDefaultAddress, "\")">(),
> ```

I don't really know what rfl does (I guess I know who to ask lol).
___
*[`include/o_rly/config/config.h` line 20 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqH7P52g0rHA7xrpFl:-OmqH7P52g0rHA7xrpFm:bebp427) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L20)):*
> ```objc
> 
>   rfl::Description<str_concat<"Bind address (default: \"", kDefaultAddress, "\")">(),
>                    std::optional<std::string>>
> ```

If this represents the parsed config, do we want to hold this in e.g. a folly::SocketAddress instead of a string (which may or may not be an actual address)?

Or at least, when this is passed around, is it assume that it's been validated?
___
*[`include/o_rly/config/config.h` line 37 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHvCPCyQOH_BOLIUw:-OmqHvCQCtT8DEUDysYm:b-kiy1ab) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L37)):*
> ```objc
> };
> 
> struct TlsCredentials {
> ```

There's a lot to TLS config besides certificates (eg. credentials).  Do you envision a larger struct encapsulating this or should we rename for future growth?
___
*[`include/o_rly/config/config.h` line 43 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHURb1D8TiuUNPGZD:-OmqHURb1D8TiuUNPGZE:btumthy) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L43)):*
> ```objc
>   rfl::Description<"Path to TLS private key file", std::optional<std::string>> key_file;
>   rfl::Description<
>       str_concat<"Skip TLS, insecure mode (default: ", bool_to_str<kDefaultInsecure>(), ")">(),
> ```

At least in moxygen, this doesn't mean "skip tls" but instead it means "use a default certificate/key that is compiled into the code"
___
*[`include/o_rly/config/config.h` line 47 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHelQ4Zkurc-XMnPr:-OmqHelQ4Zkurc-XMnPs:b-e8eeif) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L47)):*
> ```objc
>       insecure;
> 
>   bool insecureOrDefault() const { return insecure.value().value_or(kDefaultInsecure); }
> ```

Isn't another way to handle this by declaring it as `bool insecure{kDefulatInsecure}`?  or does the rfl thing require everything to be optional?
___
*[`include/o_rly/config/config.h` line 51 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqINi-EJpQH-sTIXsO:-OmqINi-EJpQH-sTIXsP:b-nwoiwm) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L51)):*
> ```objc
> 
> struct ListenerConfig {
>   static constexpr char kDefaultEndpoint[] = "/moq-relay";
> ```

std::string_view?
___
*[`include/o_rly/config/string_literal.h` line 3 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqIYXB7RsG22w-f-Nb:-OmqIYXB7RsG22w-f-Nc:b-oyvwp0) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/string_literal.h#L3)):*
> ```objc
> #pragma once
> 
> // Compile-time string utilities for building rfl::Description strings
> ```

My goodness.  This is required to make rfl happy because it has it's own string types?
___
*[`scripts/config-schema-to-markdown.sh` line 3 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqIqHW-6Em9Ywzdkv_:-OmqIqHW-6Em9Ywzdkva:bwjszr1) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/scripts/config-schema-to-markdown.sh#L3)):*
> ```Shell
> #!/usr/bin/env bash
> # Generate markdown config reference from o-rly JSON schema on stdin.
> # Usage: o_rly dump-config-schema | scripts/gen-config-reference.sh
> ```

Can you paste / link the output of this script?
___
*[`src/main.cpp` line 17 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqL9ei9_HhaF6PdbI3:-OmqL9ei9_HhaF6PdbI4:b123r9b) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/main.cpp#L17)):*
> ```cpp
> // On success returns LoadResult for "serve" mode.
> // On error returns an exit code (for early-exit subcommands or failures).
> folly::Expected<openmoq::o_rly::config::LoadResult, int> handleSubcommand(int argc, char* argv[]) {
> ```

What do you think about moving this into another file if you think there may be a lot of subcommands?  I don't want to clutter main too much?

Also I wonder if we should build a dispatch table of command -> handler rather than this cascading if.  This is ok for now though.
___
*[`src/main.cpp` line 18 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqLPd25eDOpaBtTxgL:-OmqLPd25eDOpaBtTxgM:b-6a85x1) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/main.cpp#L18)):*
> ```cpp
> // On error returns an exit code (for early-exit subcommands or failures).
> folly::Expected<openmoq::o_rly::config::LoadResult, int> handleSubcommand(int argc, char* argv[]) {
>   std::string subcommand = "serve";
> ```

Let's use symbolic constants for subcommands (kServeCommand, kValidateConfigCommand) etc
___
*[`src/main.cpp` line 29 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqMBjr6IN496yc40jW:-OmqMBjr6IN496yc40jX:b5nrafd) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/main.cpp#L29)):*
> ```cpp
> 
>   if (subcommand != "serve" && subcommand != "validate-config") {
>     std::cerr << "Unknown subcommand: " << subcommand << "\n";
> ```

I think these are probably appropriate straight uses of std::cerr rather than our logging framework.
___
*[`src/config/loader.cpp` line 14 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqIz4qBvQRolSQSKMi:-OmqIz4qBvQRolSQSKMj:b-i0fhd7) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L14)):*
> ```cpp
> namespace {
> 
> std::string readFileContents(const std::string& path) {
> ```

Unsure if you do or do not want additional folly deps, but there's helpers for things like this - see folly::File and folly::readFile (see https://github.com/facebook/folly/blob/main/folly/FileUtil.h)
___
*[`src/config/loader.cpp` line 37 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqJIWv7vqJZCU_9kI7:-OmqJIWv7vqJZCU_9kI8:blv3y5y) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L37)):*
> ```cpp
>   // Matches reflect-cpp v0.18.0 error format for NoExtraFields violations.
>   // Update if upgrading reflect-cpp changes the error message format.
>   std::regex fieldRegex(R"(Value named '([^']+)' not used)");
> ```

std::regex is notoriously slow, so we should use it cautiously.
___
*[`src/config/loader.cpp` line 64 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqJZAmDEY__fBd3T71:-OmqJZAmDEY__fBd3T72:b-ojg18h) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L64)):*
> ```cpp
>   lr.config = std::move(*result);
> 
>   auto unknownWarnings = detectUnknownFields(content);
> ```

Wait, you already called yaml::read and now detectUnknownFields calls it again.  Can you combine into a single read() call or am I missing something?
___
*[`src/config/loader.cpp` line 106 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqK6zK-wQF5rmMBCLm:-OmqK6zK-wQF5rmMBCLn:bnmxrfe) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L106)):*
> ```cpp
> 
>   // Listener must have udp configured
>   if (!listener.udp.value().has_value()) {
> ```

Again don't know rfl, but what is the type of udp() here?  If it was itself _optional_ then we don't need value().has_value() which looks silly.
___
*[`src/config/loader.cpp` line 113 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqJpJA73ds2wHg4AFo:-OmqJpJA73ds2wHg4AFp:bpm3al3) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L113)):*
> ```cpp
>     auto sock = listener.udp.value()->socketOrDefault();
>     uint16_t port = sock.portOrDefault();
>     if (port == 0) {
> ```

Eventually we will want to allow port 0 for integration tests (pick an available one), but this is ok for now.
___
*[`src/config/loader.cpp` line 147 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqK_lA11Fvm2TR4MzH:-OmqK_lA11Fvm2TR4MzI:b-kve6j7) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L147)):*
> ```cpp
> }
> 
> std::vector<std::string> warnConfig(const Config& config) {
> ```

Thoughts on combining the config validation into a single pass that adds both warnings and errors?  I think it might be easier if we only had one place in the code where we think about e.g. listener config
___
*[`src/config/loader.cpp` line 175 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqKR3F-kB8bqpqM1A7:-OmqKR3F-kB8bqpqM1A8:b-l746qg) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L175)):*
> ```cpp
> }
> 
> std::string moqtVersionsToString(const ListenerConfig& listener) {
> ```

See folly::join

michalhosna (commented):
Just a partial submission. I'll do the rest tomorrow 

> @michalhosna made 12 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29#-:-Omuvgtv9aUVPqDDMsPK:bghugkc)* status: all files reviewed, 19 unresolved discussions (waiting on afrind).

___
*[`include/o_rly/config/config.h` line 13 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqGreI-KlAQI6_sQdG:-OmsiyY12qHkruI256V9:b-575fms) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L13)):*

Do we like deeply nested namespaces? I 'm ok with it, just want to ask
</blockquote></details>

:shrug: I didn't put much thought into it; I just started with something—let's call it strawman. Happy to change either way. 
___
*[`include/o_rly/config/config.h` line 19 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqH3Av7ukfrw9doWbq:-OmsjAB_C3mAFgg0LZUM:b-sen5b9) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L19)):*
> A C++20 library for fast serialization, deserialization and validation using reflection. Supports JSON, Avro, BSON, Cap'n Proto, CBOR, CSV, flexbuffers, msgpack, parquet, TOML, UBJSON, XML, YAML / msgpack.org\[C++20\]

It actually uses yaml-cpp for parsing here. The main point is having validation and documentation in a single place. 

Given that we are trying to hit a moving target, as we are not yet sure what kind of configuration, scoping, MOQT features, filters, etc. will need what kind of documentation. Therefore, the configuration will evolve quite a bit. Having a strong documentation and validation story seemed important to me. 

Not that rfl is perfect, but the best I know about in cpp land. 
___
*[`include/o_rly/config/config.h` line 20 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqH7P52g0rHA7xrpFl:-Omun91-5mSYYqBRezBy:bdim2uq) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L20)):*

If this represents the parsed config, do we want to hold this in e.g. a folly::SocketAddress instead of a string (which may or may not be an actual address)?

Or at least, when this is passed around, is it assume that it's been validated?
</blockquote></details>

This needs to map 1:1 to the YAML file. It's not really meant to be passed around. 

It would be nice if one could have the same structure representing the config file and pass it around. But usually those two tasks require different tradeoffs. 

E.g. Folly::SocketAddress represents address and port. But we have those two as different fields in the YAML. 

We could merge them. Not a big deal here, but generally it is good to be able to decouple the config for human readability and internal representation for correctness. 

I tried to keep this PR simple, but there will 100% need to be another layer that provides validated structures with nice concrete types for passing around as the need comes. 

I am generally more a fan of implementing stuff once it's needed and not trying to do too much (over-)engineering beforehand, but I'm happy to adapt to a different style. 
___
*[`include/o_rly/config/config.h` line 37 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHvCPCyQOH_BOLIUw:-Omsl9AYCn79GY-Lzr-s:b7w1d3n) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L37)):*

There's a lot to TLS config besides certificates (eg. credentials).  Do you envision a larger struct encapsulating this or should we rename for future growth?
</blockquote></details>

I put the first name that came to mind here; renaming is cheap in the future. I do expect to rewrite this completely. 

Do you want to rename it now? Any concrete suggestions? `TlsConfig`? 
___
*[`include/o_rly/config/config.h` line 43 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHURb1D8TiuUNPGZD:-Omt4ffr9JnxaAFvjxc0:b-886gsx) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L43)):*

At least in moxygen, this doesn't mean "skip tls" but instead it means "use a default certificate/key that is compiled into the code"
</blockquote></details>

Fixed
___
*[`include/o_rly/config/config.h` line 47 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHelQ4Zkurc-XMnPr:-OmuhCTN4YK4Yl1saKhM:bd95jkg) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L47)):*

Isn't another way to handle this by declaring it as `bool insecure{kDefulatInsecure}`?  or does the rfl thing require everything to be optional?
</blockquote></details>

The std::optional wrapper is what tells reflect-cpp the field can be omitted from the YAML. 
Changing to bool would make the field required regardless of the initializer, as reflect-cpp would just see the bool type. 
___
*[`include/o_rly/config/config.h` line 51 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqINi-EJpQH-sTIXsO:-OmutH1U8PIaCPi3wuAl:b-wfkgx5) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L51)):*

std::string_view?
</blockquote></details>

That won't work with rfl for the compile-time string in the documentation. 

Starting to get feeling you won't like rfl much. But I don't think keeping documentation separate and updating it manually will be better. 
___
*[`include/o_rly/config/string_literal.h` line 3 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqIYXB7RsG22w-f-Nb:-OmursM96kTsFsA8cW3U:b-52gjvm) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/string_literal.h#L3)):*

My goodness.  This is required to make rfl happy because it has it's own string types?
</blockquote></details>

It's only so I can concatenate strings with constants. 
I want to print defaults in the description and have them defined only once so the documentation cannot drift. 

It's gnarly, but as I expect the configuration file itself will get gnarly down the line, I think it's a fair tradeoff to have this. 
___
*[`scripts/config-schema-to-markdown.sh` line 3 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqIqHW-6Em9Ywzdkv_:-Omt15yZ13N-xp77O75N:b-463qf3) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/scripts/config-schema-to-markdown.sh#L3)):*
The JSON scheme can be consumed by multiple tools; this is just a vibe-coded helper to have something humanly readable in-tree 

-------

# o-rly Configuration Reference

> Auto-generated from JSON schema. Do not edit manually.

## cache

Relay cache settings (default: true)

| Field | Type | Description |
| --- | --- | --- |
| `enabled` | boolean | Enable relay cache (default: true) |
| `max_groups_per_track` | integer | Max cached groups per track, ignored when disabled (default: 3) |
| `max_tracks` | integer | Max cached tracks, ignored when disabled (default: 100) |

## listeners

Listener definitions (currently exactly one supported)

| Field | Type | Description |
| --- | --- | --- |
| `endpoint` | string | WebTransport endpoint path (default: "/moq-relay") |
| `moqt_versions` | integer[] | MOQT draft versions (empty = all supported) |
| `name` | string | Listener name |
| `tls_credentials.cert_file` | string | Path to TLS certificate file |
| `tls_credentials.insecure` | boolean | Insecure mode, use default compiled-in certificate (default: false) |
| `tls_credentials.key_file` | string | Path to TLS private key file |
| `udp.socket.address` | string | Bind address (default: "::") |
| `udp.socket.port` | integer | Listen port, 1-65535 (default: 9668) |
___
*[`src/main.cpp` line 29 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqMBjr6IN496yc40jW:-Omt1eY738x85N8_Gzvj:b-9tjfe2) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/main.cpp#L29)):*

I think these are probably appropriate straight uses of std::cerr rather than our logging framework.
</blockquote></details>

Yep

FYI You can mark comments as informational in reviewable, so that I have to ack them, but it's clear its just a FYI. 

___
*[`src/config/loader.cpp` line 113 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqJpJA73ds2wHg4AFo:-Omuscb21vHdYsFerPJq:b-hwwqpq) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L113)):*

Eventually we will want to allow port 0 for integration tests (pick an available one), but this is ok for now.
</blockquote></details>

We'll then need a way to get the assigned port out, I expect. I would rather not deal with that now. Or does Moxigen Folly already solve that?

afrind (commented):
> @afrind reviewed all commit messages, made 8 comments, and resolved 10 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29)* status: all files reviewed, 9 unresolved discussions (waiting on michalhosna).

___
*[`include/o_rly/config/config.h` line 13 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqGreI-KlAQI6_sQdG:-Omv0sA1D0qCyJDLRP1q:bxtr4zt) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L13)):*

:shrug: I didn't put much thought into it; I just started with something—let's call it strawman. Happy to change either way.
</blockquote></details>

Keep for now
___
*[`include/o_rly/config/config.h` line 19 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqH3Av7ukfrw9doWbq:-Omv1uel2OKjEMsg62-K:bv8v9w6) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L19)):*

> A C++20 library for fast serialization, deserialization and validation using reflection. Supports JSON, Avro, BSON, Cap'n Proto, CBOR, CSV, flexbuffers, msgpack, parquet, TOML, UBJSON, XML, YAML / msgpack.org\[C++20\]

It actually uses yaml-cpp for parsing here. The main point is having validation and documentation in a single place. 

Given that we are trying to hit a moving target, as we are not yet sure what kind of configuration, scoping, MOQT features, filters, etc. will need what kind of documentation. Therefore, the configuration will evolve quite a bit. Having a strong documentation and validation story seemed important to me. 

Not that rfl is perfect, but the best I know about in cpp land.
</blockquote></details>

Ok, seems like it might be heavyweight but it's fine.  Let's go.
___
*[`include/o_rly/config/config.h` line 20 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqH7P52g0rHA7xrpFl:-Omv279FEIS4wCPa6RXi:b-rkyyh8) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L20)):*

This needs to map 1:1 to the YAML file. It's not really meant to be passed around. 

It would be nice if one could have the same structure representing the config file and pass it around. But usually those two tasks require different tradeoffs. 

E.g. Folly::SocketAddress represents address and port. But we have those two as different fields in the YAML. 

We could merge them. Not a big deal here, but generally it is good to be able to decouple the config for human readability and internal representation for correctness. 

I tried to keep this PR simple, but there will 100% need to be another layer that provides validated structures with nice concrete types for passing around as the need comes. 

I am generally more a fan of implementing stuff once it's needed and not trying to do too much (over-)engineering beforehand, but I'm happy to adapt to a different style.
</blockquote></details>

There's [folly::IPAddress](https://github.com/facebook/folly/blob/main/folly/IPAddress.h) too.  But string works for now -- unclear if we want non-IP addresses (eg unix sock?)
___
*[`include/o_rly/config/config.h` line 37 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHvCPCyQOH_BOLIUw:-Omv2U5m1tn05RLtzf1O:b-jrln6o) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L37)):*

I put the first name that came to mind here; renaming is cheap in the future. I do expect to rewrite this completely. 

Do you want to rename it now? Any concrete suggestions? `TlsConfig`?
</blockquote></details>

Kind of like TLSConfig better, but not blocking on this.
___
*[`include/o_rly/config/config.h` line 47 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHelQ4Zkurc-XMnPr:-Omv2flJBIcG0HhZNCyC:bxuq5hd) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L47)):*

The std::optional wrapper is what tells reflect-cpp the field can be omitted from the YAML. 
Changing to bool would make the field required regardless of the initializer, as reflect-cpp would just see the bool type.
</blockquote></details>

Hmm, that's unfortunate -- we want things optional _in the config_ but always have a value _in the structure_.

We could make a macro like 
```
#define ORLY_CONFIG(DESC, TYPE, NAME, DEFAULT) \
  rfl::Description<DESC, std::optional<TYPE>> _ ## NAME; \
  TYPE NAME() { return _ ## NAME.value_or(DEFAULT); }
```

or both `ORLY_OPTIONAL_CONFIG` (with default) and `ORLY_REQUIRED_CONFIG` (without)
___
*[`include/o_rly/config/config.h` line 51 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqINi-EJpQH-sTIXsO:-Omv3sSz0dJqUQqoVhpX:b6sj3xe) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L51)):*

That won't work with rfl for the compile-time string in the documentation. 

Starting to get feeling you won't like rfl much. But I don't think keeping documentation separate and updating it manually will be better.
</blockquote></details>

Ah, I see rfl requires this specific type.
___
*[`include/o_rly/config/string_literal.h` line 3 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqIYXB7RsG22w-f-Nb:-Omv64J38t5SPiP1COtN:b-b41r2y) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/string_literal.h#L3)):*

It's only so I can concatenate strings with constants. 
I want to print defaults in the description and have them defined only once so the documentation cannot drift. 

It's gnarly, but as I expect the configuration file itself will get gnarly down the line, I think it's a fair tradeoff to have this.
</blockquote></details>

I'm going to look away
___
*[`src/config/loader.cpp` line 113 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqJpJA73ds2wHg4AFo:-Omv6GhA330IN74yYYCt:b-uvif2p) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L113)):*

We'll then need a way to get the assigned port out, I expect. I would rather not deal with that now. Or does Moxigen Folly already solve that?
</blockquote></details>

In some integration tests that launch the binary independently, we [don't laugh] print it to stdout.

But we also have a framework where you can run our proxy as a thread inside another program.  This is useful as an C++ integration  test that has the clients in some threads, servers in another, and proxy is right there -- you get the bound address with a getAddress API.

michalhosna (commented):
I split off ParsedConfig to represent the structures required for file parsing and config that represents parsed and validated config. 
I should have done this from the start. 

It's quite a different approach, so it may be wise to not review the diff to the previous revision of this PR, which is the default, but review the diff to main. Time to flex reviewable features. 

Also, I rebased this on main with different formatting. With reviewable, this should not clutter the diff much, 

> @michalhosna made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29#-:-OmywYWb2kXle3yWkTeV:bcimxkh)* status: 1 of 16 files reviewed, 9 unresolved discussions (waiting on afrind and mondain).

___
*[`include/o_rly/config/config.h` line 20 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqH7P52g0rHA7xrpFl:-Omyw6UE1TJDDmReMRVh:b-fryda7) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L20)):*

There's [folly::IPAddress](https://github.com/facebook/folly/blob/main/folly/IPAddress.h) too.  But string works for now -- unclear if we want non-IP addresses (eg unix sock?)
</blockquote></details>

See the new file structure for folly::IPAddress. 

Does QUIC over unix sockets make sense? I am not sure, probably not? So that would be raw moqt over unix sockets? What about datagrams? 

I had this line of thought and decided that it doesn't need resolving now :smile: 
___
*[`include/o_rly/config/config.h` line 37 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHvCPCyQOH_BOLIUw:-OmzxPLA91wYFhTZD3wn:bgc9via) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L37)):*

Kind of like TLSConfig better, but not blocking on this.
</blockquote></details>

Changed to TLS config and just `tls` in the yaml. 
___
*[`include/o_rly/config/config.h` line 47 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHelQ4Zkurc-XMnPr:-OmyvRsc79mnDoaFkAcb:bs7b1ae) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L47)):*

Hmm, that's unfortunate -- we want things optional _in the config_ but always have a value _in the structure_.

We could make a macro like 
```
#define ORLY_CONFIG(DESC, TYPE, NAME, DEFAULT) \
  rfl::Description<DESC, std::optional<TYPE>> _ ## NAME; \
  TYPE NAME() { return _ ## NAME.value_or(DEFAULT); }
```

or both `ORLY_OPTIONAL_CONFIG` (with default) and `ORLY_REQUIRED_CONFIG` (without)
</blockquote></details>

What about optional fields that doesn't have a default? Which I expect would be the majority. 

And do the macros even help? 

```
 rfl::Description<"Path to TLS private key file", std::optional<std::string>> key_file;
```
vs 
```
 ORLY_OPTIONAL_CONFIG_NO_DEFAULT("Path to TLS private key file", std::optional<std::string>>, key_file);
```

I think the explicitness and wordiness here is actually better then having a level of indirection. 

What do you think with the current split of concerns between `Config` and `ParsedConfig`? 
___
*[`src/main.cpp` line 17 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqL9ei9_HhaF6PdbI3:-OmypxYQ2p3tSk6yb5UP:b-zf8m96) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/main.cpp#L17)):*

What do you think about moving this into another file if you think there may be a lot of subcommands?  I don't want to clutter main too much?

Also I wonder if we should build a dispatch table of command -> handler rather than this cascading if.  This is ok for now though.
</blockquote></details>

I am not sure if we plan to have a lot of subcommands. 

Currently its like, "serve" = please continue in the main. Other commands = quick exit. 

Using a dispatch table in a different file breaks this "please continue". Should the rest of the main be moved into some kind of serve handler and make the main essentially empty? It makes sense, but also it's maybe too drastic given that that's the main way. 

Happy to do it if you thinks its better. I wasn't even sure if subcommands are the way to go. Is a serve good default name? 
___
*[`src/main.cpp` line 18 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqLPd25eDOpaBtTxgL:-Omyufal4NZMfztSNTS5:b-896fix) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/main.cpp#L18)):*

Let's use symbolic constants for subcommands (kServeCommand, kValidateConfigCommand) etc
</blockquote></details>

Done.
___
*[`src/config/loader.cpp` line 14 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqIz4qBvQRolSQSKMi:-OmzxwapB-CXdau5Curg:b-896fix) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L14)):*

Unsure if you do or do not want additional folly deps, but there's helpers for things like this - see folly::File and folly::readFile (see https://github.com/facebook/folly/blob/main/folly/FileUtil.h)
</blockquote></details>

Done.
___
*[`src/config/loader.cpp` line 37 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqJIWv7vqJZCU_9kI7:-On--zPX-WkL3T0VKqN6:b-3qule6) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L37)):*

std::regex is notoriously slow, so we should use it cautiously.
</blockquote></details>

Dropped this whole code for now
___
*[`src/config/loader.cpp` line 64 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqJZAmDEY__fBd3T71:-Omzxy-h3lLhcvt0fbX6:b-oizvi5) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L64)):*

Wait, you already called yaml::read and now detectUnknownFields calls it again.  Can you combine into a single read() call or am I missing something?
</blockquote></details>

I wanted to always warn on unknown fields and only hard error in strict mode. The RFL doesn't support this. Its either ignore or hard error. Most libraries don't support it, given that you need a wrapper type for that and warnings don't have a clear structure. 

But maybe I am too cautious. I simplified it to strict mode with hard error and ignore during non strict. Given that we provide the validation subcommand, it should be fine. 
___
*[`src/config/loader.cpp` line 106 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqK6zK-wQF5rmMBCLm:-OmzySyvEtqaLN5FeX25:b2awor9) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L106)):*

Again don't know rfl, but what is the type of udp() here?  If it was itself _optional_ then we don't need value().has_value() which looks silly.
</blockquote></details>

The type of UDP is rfl::Description. 

This really is only meant as a "parser generator". 
___
*[`src/config/loader.cpp` line 113 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqJpJA73ds2wHg4AFo:-OmyumgBEdOULL2kcDgd:by3xbey) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L113)):*
> But we also have a framework where you can run our proxy as a thread inside another program.

Then again, this is for parsing, not passing around. 

I am going to do an intermediate config structure so this is clearer. 
___
*[`src/config/loader.cpp` line 147 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqK_lA11Fvm2TR4MzH:-OmzywlT16v1IsWfA5Dq:bfgnvvv) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L147)):*

Thoughts on combining the config validation into a single pass that adds both warnings and errors?  I think it might be easier if we only had one place in the code where we think about e.g. listener config
</blockquote></details>

See comment above https://reviewable.io/reviews/openmoq/o-rly/29#-OmqJZAmDEY__fBd3T71
___
*[`src/config/loader.cpp` line 175 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqKR3F-kB8bqpqM1A7:-OmzzO8OB8yXCSaGv9Nn:b-896fix) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/config/loader.cpp#L175)):*

See folly::join
</blockquote></details>

Done.

afrind (changes requested):
> @afrind reviewed 15 files and all commit messages, made 7 comments, and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29)* status: all files reviewed, 4 unresolved discussions (waiting on michalhosna).

___
*[`include/o_rly/config/config.h` line 20 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqH7P52g0rHA7xrpFl:-On-JwwV2I7WEDOTbA2a:b-k2e9qu) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L20)):*

See the new file structure for folly::IPAddress. 

Does QUIC over unix sockets make sense? I am not sure, probably not? So that would be raw moqt over unix sockets? What about datagrams? 

I had this line of thought and decided that it doesn't need resolving now :smile:
</blockquote></details>

we'll have a tcp fallback for moqt soon (see QMUX)
___
*[`include/o_rly/config/config.h` line 47 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHelQ4Zkurc-XMnPr:-On-J_KZ26IthsrmDrLL:bkn6027) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L47)):*

What about optional fields that doesn't have a default? Which I expect would be the majority. 

And do the macros even help? 

```
 rfl::Description<"Path to TLS private key file", std::optional<std::string>> key_file;
```
vs 
```
 ORLY_OPTIONAL_CONFIG_NO_DEFAULT("Path to TLS private key file", std::optional<std::string>>, key_file);
```

I think the explicitness and wordiness here is actually better then having a level of indirection. 

What do you think with the current split of concerns between `Config` and `ParsedConfig`?
</blockquote></details>

No they key difference is that the std::optional<> piece is added automatically by the macro _if_ the config is optional.  So it would be:

`ORLY_OPTIONAL_CONFIG(`"Path to TLS private key file", std::string, key_file, "")`
___
*[`src/main.cpp` line 17 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqL9ei9_HhaF6PdbI3:-On-MKBCD8adbuiqS7Mr:bv8lf0h) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/main.cpp#L17)):*

I am not sure if we plan to have a lot of subcommands. 

Currently its like, "serve" = please continue in the main. Other commands = quick exit. 

Using a dispatch table in a different file breaks this "please continue". Should the rest of the main be moved into some kind of serve handler and make the main essentially empty? It makes sense, but also it's maybe too drastic given that that's the main way. 

Happy to do it if you thinks its better. I wasn't even sure if subcommands are the way to go. Is a serve good default name?
</blockquote></details>

I'd rather move this out to the config library and keep main clean.  `processConfigCommand(command)`

For the "default" command -- pass empty string? 
___
*[`CMakeLists.txt` line 173 at r2](https://reviewable.io/reviews/openmoq/o-rly/29#-On-JKav66Q140RseFmR:-On-JKav66Q140RseFmS:bpt8a8a) ([raw file](https://github.com/openmoq/o-rly/blob/d9418115fa872dfd95f5df15add6a2a2f8b32745/CMakeLists.txt#L173)):*
> ```CMake
>   # override the feature so CMake doesn't warn about mixed link strategies.
>   set_property(TARGET o_rly_config_test PROPERTY
>     LINK_LIBRARY_OVERRIDE "WHOLE_ARCHIVE,gflags_nothreads_static"
> ```

Ew. Can we file an issue in the appropriate place here?  Is o-rly still using getdeps?
___
*[`src/main.cpp` line 83 at r2](https://reviewable.io/reviews/openmoq/o-rly/29#-On-N8XR554Aw6_nd15k:-On-N8XR554Aw6_nd15l:b1ewn5f) ([raw file](https://github.com/openmoq/o-rly/blob/d9418115fa872dfd95f5df15add6a2a2f8b32745/src/main.cpp#L83)):*
> ```cpp
>   const auto& cache = resolved.cache;
> 
>   return std::visit(
> ```

I don't love this use of variant.  But I don't think this is the long term play so I can live with it in the name of progress.
___
*[`src/config/config_resolver.cpp` line 95 at r2](https://reviewable.io/reviews/openmoq/o-rly/29#-On-KG_SAcjiHJ2_3IEV:-On-KG_TB_D1WfsskW06:b-uid5iv) ([raw file](https://github.com/openmoq/o-rly/blob/d9418115fa872dfd95f5df15add6a2a2f8b32745/src/config/config_resolver.cpp#L95)):*
> ```cpp
> }
> 
> Config resolveConfig(const ParsedConfig& config) {
> ```

This creates boilerplate that is easily forgotten and a pain in the butt.  e.g. I added my config struct but not the file, or the file but not the struct, and missed this key step.  Ask AI if there's a way to autogenerate both the Config struct itself and this function?
___
*[`tests/config/test_utils.h` line 12 at r2](https://reviewable.io/reviews/openmoq/o-rly/29#-On-LMZC-Y9_NIO4wb_H:-On-LMZC-Y9_NIO4wb_I:b-jyuipf) ([raw file](https://github.com/openmoq/o-rly/blob/d9418115fa872dfd95f5df15add6a2a2f8b32745/tests/config/test_utils.h#L12)):*
> ```objc
> 
> // RAII helper: writes YAML content to a unique temp file, removes it on destruction.
> class TempYamlFile {
> ```

Pretty sure there's a folly helper for this too...BUT it currently uses boost::filesystem (BOO).

michalhosna (commented):
> @michalhosna made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29)* status: all files reviewed, 4 unresolved discussions (waiting on afrind).

___
*[`CMakeLists.txt` line 173 at r2](https://reviewable.io/reviews/openmoq/o-rly/29#-On-JKav66Q140RseFmR:-On22w0Y5cTaHqE3C9Jr:btipurj) ([raw file](https://github.com/openmoq/o-rly/blob/d9418115fa872dfd95f5df15add6a2a2f8b32745/CMakeLists.txt#L173)):*

Ew. Can we file an issue in the appropriate place here?  Is o-rly still using getdeps?
</blockquote></details>

Honestly, I am completely lost in the current state of the build system of the dependencies. I am happy to have a project that builds, and I really don't want to have to go dig given the time crunch.

It builds; it doesn't pollute the compiler output with new garbage. I despise C++ ecosystem standard of having a wall of warnings during the compile that everyone ignores. 

I am happy to accept any suggestions on how to solve this better. I wasn't even sure if I should for the new dependencies here use FetchContent, submodules, different way. H

Honestly, I really don't care that much. Get me a script that builds and doesn't make me look at C++'s horrible deps & build ecosystem again :smile: (just to be clear, I feel the same way with python, but they got `uv` now)
___
*[`include/o_rly/config/config.h` line 47 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHelQ4Zkurc-XMnPr:-On2-jB56u_JyNQmy1fp:b-86b2k2) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L47)):*

No they key difference is that the std::optional<> piece is added automatically by the macro _if_ the config is optional.  So it would be:

`ORLY_OPTIONAL_CONFIG(`"Path to TLS private key file", std::string, key_file, "")`
</blockquote></details>

My mistake, I always was thinking about. 
```
 rfl::Description<"Path to TLS private key file", std::optional<std::string>> key_file;
 ORLY_OPTIONAL_CONFIG_NO_DEFAULT("Path to TLS private key file", std::string>, key_file);
```

My problem with your suggestion is that `""` default is not the same as not having a default. I generally want to know default vs explicit set down the line. 
___
*[`src/main.cpp` line 17 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqL9ei9_HhaF6PdbI3:-On24NLGBrpjuW7YyIli:bhn3dur) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/main.cpp#L17)):*

I'd rather move this out to the config library and keep main clean.  `processConfigCommand(command)`

For the "default" command -- pass empty string?
</blockquote></details>

Now I am lost. What do you actually suggest/want? 
___
*[`src/config/config_resolver.cpp` line 95 at r2](https://reviewable.io/reviews/openmoq/o-rly/29#-On-KG_SAcjiHJ2_3IEV:-On20JDk1IKhJ9uef_ot:bsd4k7c) ([raw file](https://github.com/openmoq/o-rly/blob/d9418115fa872dfd95f5df15add6a2a2f8b32745/src/config/config_resolver.cpp#L95)):*
I don't think this is a boilerplate. And I don't think there can be a way to autogenerate. 

You need to express your config constraints somewhere. Config constraints can be fairly complicated, and you need to be able to validate and express complicated stuff.

For example, if we have multiple listeners that are named, then services can limit on which listeners they are available. You have to be able to check that the services list valid listeners. This must be a step after the parsing, and its something that cannot be reasonably expressed in something like JSON schema or other config DSL. 

> I added my config struct but not the file, or the file but not the struct, and missed this key step

The type system should have helped you not miss this step. Your newly added structure will always be missing on one side. 

If you added only to the parser and not to the `Config`, you simply won't have the field in the application. You should be able to notice that. 

If you add only to Config, and forget here, it won't compile. 

-----

I am used to this pattern of two pass config loading. First parse, and do checks that can be easily done at that step (required fields, required format of those fields, allowed keys etc.) and then have a somewhat manually crafted mapping function (this one) that does the second level of validation and also crafts nice types and structures. 

Do you have a different pattern you can recommend?

michalhosna (commented):
I added the docs, dropped defaults, and reorganized the code.  

> @michalhosna made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29#-:-On4-xDe2tSQHHTA3kWK:b-frmxt5)* status: 6 of 22 files reviewed, 4 unresolved discussions (waiting on afrind).

___
*[`include/o_rly/config/config.h` line 47 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHelQ4Zkurc-XMnPr:-On3s1dv06dq3BDKvDgE:byv1kon) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L47)):*

My mistake, I always was thinking about. 
```
 rfl::Description<"Path to TLS private key file", std::optional<std::string>> key_file;
 ORLY_OPTIONAL_CONFIG_NO_DEFAULT("Path to TLS private key file", std::string>, key_file);
```

My problem with your suggestion is that `""` default is not the same as not having a default. I generally want to know default vs explicit set down the line.
</blockquote></details>

I created a separate PR to show the diff: https://github.com/openmoq/o-rly/pull/36
___
*[`src/main.cpp` line 17 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqL9ei9_HhaF6PdbI3:-On401mH6tsgcfU8YdJb:b-so7m3b) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/src/main.cpp#L17)):*

Now I am lost. What do you actually suggest/want?
</blockquote></details>

Discussed elsewhere, I moved the config subcommands to the config tree. And kept the rest as a default here.

afrind (approved):
> @afrind partially reviewed 16 files and all commit messages, made 5 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29)* status: all files reviewed, 1 unresolved discussion (waiting on michalhosna).

___
*[`CMakeLists.txt` line 173 at r2](https://reviewable.io/reviews/openmoq/o-rly/29#-On-JKav66Q140RseFmR:-On8OzQW2F21-YwE2egP:b-b4gvho) ([raw file](https://github.com/openmoq/o-rly/blob/d9418115fa872dfd95f5df15add6a2a2f8b32745/CMakeLists.txt#L173)):*

Honestly, I am completely lost in the current state of the build system of the dependencies. I am happy to have a project that builds, and I really don't want to have to go dig given the time crunch.

It builds; it doesn't pollute the compiler output with new garbage. I despise C++ ecosystem standard of having a wall of warnings during the compile that everyone ignores. 

I am happy to accept any suggestions on how to solve this better. I wasn't even sure if I should for the new dependencies here use FetchContent, submodules, different way. H

Honestly, I really don't care that much. Get me a script that builds and doesn't make me look at C++'s horrible deps & build ecosystem again :smile: (just to be clear, I feel the same way with python, but they got `uv` now)
</blockquote></details>

Please file an issue _somewhere_ so we don't forget to clean this up.
___
*[`include/o_rly/config/config.h` line 47 at r1](https://reviewable.io/reviews/openmoq/o-rly/29#-OmqHelQ4Zkurc-XMnPr:-On8PCDM2Fxi4k77Rcwu:b-jqaub8) ([raw file](https://github.com/openmoq/o-rly/blob/e8a87d3053272bfe762761acc71569c1dfe68137/include/o_rly/config/config.h#L47)):*
> know default vs explicit set down the line.

You mean it matters if someone sets port=443 vs not setting port, even though the app sees port=443 in either case?  I'm not sure why, but I'm not lying in the road here.
___
*[`src/config/config_resolver.cpp` line 95 at r2](https://reviewable.io/reviews/openmoq/o-rly/29#-On-KG_SAcjiHJ2_3IEV:-On8PW4j0WpVUmfmUds9:byi94u8) ([raw file](https://github.com/openmoq/o-rly/blob/d9418115fa872dfd95f5df15add6a2a2f8b32745/src/config/config_resolver.cpp#L95)):*

I don't think this is a boilerplate. And I don't think there can be a way to autogenerate. 

You need to express your config constraints somewhere. Config constraints can be fairly complicated, and you need to be able to validate and express complicated stuff.

For example, if we have multiple listeners that are named, then services can limit on which listeners they are available. You have to be able to check that the services list valid listeners. This must be a step after the parsing, and its something that cannot be reasonably expressed in something like JSON schema or other config DSL. 

> I added my config struct but not the file, or the file but not the struct, and missed this key step

The type system should have helped you not miss this step. Your newly added structure will always be missing on one side. 

If you added only to the parser and not to the `Config`, you simply won't have the field in the application. You should be able to notice that. 

If you add only to Config, and forget here, it won't compile. 

-----

I am used to this pattern of two pass config loading. First parse, and do checks that can be easily done at that step (required fields, required format of those fields, allowed keys etc.) and then have a somewhat manually crafted mapping function (this one) that does the second level of validation and also crafts nice types and structures. 

Do you have a different pattern you can recommend?
</blockquote></details>

There are downsides to this approach (have to change 3 things: two structs, and resolve function; plus validation).  It's a lot for simple configs, but for others that have meaningful validation perhaps it's not that much more.  My goal would be to touch things as few times as possible, but I think the upsides of your approach are fine.  Let's roll.
___
*[`src/main.cpp` line 67 at r3](https://reviewable.io/reviews/openmoq/o-rly/29#-On8QT4Z4Snbl8jlxACO:-On8QT4Z4Snbl8jlxACP:b-riz6by) ([raw file](https://github.com/openmoq/o-rly/blob/5f9731a124a3d67f22c27c6c20dcede96ba11c03/src/main.cpp#L67)):*
> ```cpp
>   }
> 
>   auto result = cfg::handleConfigSubcommand(subcommand, FLAGS_config, FLAGS_strict_config, argv[0]);
> ```

Just to be clear, serve as a config subcommand loads the config?  Might be good to have a comment.
___
*[`src/config/config_resolver.cpp` line 77 at r3](https://reviewable.io/reviews/openmoq/o-rly/29#-On8Qtkr0QCv7L-4_IJX:-On8Qtkr0QCv7L-4_IJY:b-38m8v4) ([raw file](https://github.com/openmoq/o-rly/blob/5f9731a124a3d67f22c27c6c20dcede96ba11c03/src/config/config_resolver.cpp#L77)):*
> ```cpp
> 
>   if (!errors.empty()) {
>     std::ostringstream oss;
> ```

I think you could folly::join here

michalhosna (commented):
> @michalhosna made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29)* status: 21 of 22 files reviewed, all discussions resolved (waiting on afrind).

___
*[`CMakeLists.txt` line 173 at r2](https://reviewable.io/reviews/openmoq/o-rly/29#-On-JKav66Q140RseFmR:-OnHEOz4-XoXp-USMg93:bv5kq7p) ([raw file](https://github.com/openmoq/o-rly/blob/d9418115fa872dfd95f5df15add6a2a2f8b32745/CMakeLists.txt#L173)):*

Please file an issue _somewhere_ so we don't forget to clean this up.
</blockquote></details>

https://github.com/openmoq/o-rly/issues/41

michalhosna (commented):
> @michalhosna reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/29)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #34, opened 2026-03-05
Add basic HTTP admin server

afrind:
Introduces AdminServer wrapping proxygen::ScopedHTTPServer with a simple method+path route table and a generic request aggregator. Routes are registered via addRoute() before start(); async handlers can launch coroutines and respond via downstream.

We will migrate to proxygen::coro::HTTPServer when proxygen::coro finishes WebTransport support.  For now, we'll use proxygen::HTTPServer.

Adds registerBuiltinRoutes() with GET /info returning a JSON version string. Wires up --admin_port flag (default 9669, 0 to disable) in main.

Workaround: wangle::wangle_acceptor_acceptor_core omits Folly::folly_io_async_fdsock_async_fd_socket from its cmake deps (FizzAcceptorHandshakeHelper uses AsyncFdSocket); added explicitly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/34)

### DIFF
--- a/src/admin/AdminServer.cpp
+++ b/src/admin/AdminServer.cpp
@@ -1,0 +1,112 @@
+#include <o_rly/admin/AdminServer.h>
+
+#include <folly/io/IOBufQueue.h>
+#include <folly/logging/xlog.h>
+#include <proxygen/httpserver/RequestHandler.h>
+#include <proxygen/httpserver/RequestHandlerFactory.h>
+#include <proxygen/httpserver/ResponseBuilder.h>
+#include <proxygen/httpserver/ScopedHTTPServer.h>
+
+namespace openmoq::o_rly::admin {
+
+// ---------------------------------------------------------------------------
+// AdminRequestHandler — generic request aggregator
+//
+// The only proxygen::RequestHandler subclass needed. Accumulates request
+// headers and body, then dispatches to the RouteHandler on EOM.
+// ---------------------------------------------------------------------------
+class AdminRequestHandler : public proxygen::RequestHandler {
+public:
+  explicit AdminRequestHandler(RouteHandler handler) : handler_(std::move(handler)) {}
+
+  void onRequest(std::unique_ptr<proxygen::HTTPMessage> headers) noexcept override {
+    req_ = std::move(headers);
+  }
+
+  void onBody(std::unique_ptr<folly::IOBuf> body) noexcept override {
+    body_.append(std::move(body));
+  }
+
+  void onEOM() noexcept override { handler_(std::move(req_), body_.move(), downstream_); }
+
+  void onUpgrade(proxygen::UpgradeProtocol /*proto*/) noexcept override {}
+
+  void requestComplete() noexcept override { delete this; }
+
+  void onError(proxygen::ProxygenError /*err*/) noexcept override { delete this; }
+
+private:
+  RouteHandler handler_;
+  std::unique_ptr<proxygen::HTTPMessage> req_;
+  folly::IOBufQueue body_{folly::IOBufQueue::cacheChainLength()};
+};
+
+// ---------------------------------------------------------------------------
+// AdminHandlerFactory — holds the route table and creates AdminRequestHandlers
+// ---------------------------------------------------------------------------
+class AdminHandlerFactory : public proxygen::RequestHandlerFactory {
+public:
+  explicit AdminHandlerFactory(std::vector<Route> routes) : routes_(std::move(routes)) {}
+
+  void onServerStart(folly::EventBase* /*evb*/) noexcept override {}
+  void onServerStop() noexcept override {}
+
+  proxygen::RequestHandler*
+  onRequest(proxygen::RequestHandler* /*upstream*/, proxygen::HTTPMessage* msg) noexcept override {
+    const auto& method = msg->getMethodString();
+    const auto& path = msg->getPath();
+
+    for (const auto& route : routes_) {
+      if (route.method == method && route.path == path) {
+        return new AdminRequestHandler(route.handler);
+      }
+    }
+
+    // Unknown route → 404
+    return new AdminRequestHandler([](auto /*req*/, auto /*body*/, auto* downstream) {
+      proxygen::ResponseBuilder(downstream)
+          .status(404, "Not Found")
+          .body(folly::IOBuf::copyBuffer("Not Found\n"))
+          .sendWithEOM();
+    });
+  }
+
+private:
+  std::vector<Route> routes_;
+};
+
+// ---------------------------------------------------------------------------
+// AdminServer
+// ---------------------------------------------------------------------------
+
+AdminServer::AdminServer() = default;
+AdminServer::~AdminServer() = default;
+
+void AdminServer::addRoute(std::string method, std::string path, RouteHandler handler) {
+  routes_.push_back({std::move(method), std::move(path), std::move(handler)});
+}
+
+bool AdminServer::start(uint16_t port) {
+  proxygen::HTTPServerOptions options;
+  options.threads = 1;
+  options.handlerFactories.push_back(std::make_unique<AdminHandlerFactory>(routes_));
+
+  proxygen::HTTPServer::IPConfig cfg{
+      folly::SocketAddress("::", port),
+      proxygen::HTTPServer::Protocol::HTTP
+  };
+
+  try {
+    httpServer_ = proxygen::ScopedHTTPServer::start(std::move(cfg), std::move(options));
+    return true;
+  } catch (const std::exception& ex) {
+    XLOG(ERR) << "AdminServer failed to start on port " << port << ": " << ex.what();
+    return false;
+  }
+}
+
+void AdminServer::stop() {
+  httpServer_.reset();
+}
+
+} // namespace openmoq::o_rly::admin
--- a/src/admin/AdminServer.cpp
+++ b/src/admin/AdminServer.cpp
@@ -6,6 +6,7 @@
 #include <proxygen/httpserver/RequestHandlerFactory.h>
 #include <proxygen/httpserver/ResponseBuilder.h>
 #include <proxygen/httpserver/ScopedHTTPServer.h>
+#include <proxygen/lib/http/HTTPMessage.h>
 
 namespace openmoq::o_rly::admin {
 
@@ -65,7 +66,7 @@ public:
     // Unknown route → 404
     return new AdminRequestHandler([](auto /*req*/, auto /*body*/, auto* downstream) {
       proxygen::ResponseBuilder(downstream)
-          .status(404, "Not Found")
+          .status(404, proxygen::HTTPMessage::getDefaultReason(404))
           .body(folly::IOBuf::copyBuffer("Not Found\n"))
           .sendWithEOM();
     });
@@ -83,12 +84,15 @@ AdminServer::AdminServer() = default;
 AdminServer::~AdminServer() = default;
 
 void AdminServer::addRoute(std::string method, std::string path, RouteHandler handler) {
+  XCHECK(!httpServer_) << "addRoute() called after start()";
   routes_.push_back({std::move(method), std::move(path), std::move(handler)});
 }
 
+static constexpr int kAdminServerThreads = 1;
+
 bool AdminServer::start(uint16_t port) {
   proxygen::HTTPServerOptions options;
-  options.threads = 1;
+  options.threads = kAdminServerThreads;
   options.handlerFactories.push_back(std::make_unique<AdminHandlerFactory>(routes_));
 
   proxygen::HTTPServer::IPConfig cfg{
--- a/src/admin/BuiltinRoutes.cpp
+++ b/src/admin/BuiltinRoutes.cpp
@@ -1,0 +1,20 @@
+#include <o_rly/admin/BuiltinRoutes.h>
+
+#include <folly/io/IOBuf.h>
+#include <proxygen/httpserver/ResponseBuilder.h>
+
+#include <o_rly/admin/AdminServer.h>
+
+namespace openmoq::o_rly::admin {
+
+void registerBuiltinRoutes(AdminServer& server) {
+  server.addRoute("GET", "/info", [](auto /*req*/, auto /*body*/, auto* downstream) {
+    proxygen::ResponseBuilder(downstream)
+        .status(200, "OK")
+        .header("Content-Type", "application/json")
+        .body(folly::IOBuf::copyBuffer("{\"service\":\"o-rly\",\"version\":\"0.1.0\"}\n"))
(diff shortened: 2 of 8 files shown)


michalhosna (changes requested):
It's a good start, I have a couple of minor nits and a couple of blocking discussions. 

> @michalhosna reviewed 6 files and all commit messages, and made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/34#-:-On1zwb1-9O1_KdmO64T:b-ugz9w5)* status: all files reviewed, 7 unresolved discussions (waiting on afrind).

___
*[a discussion](https://reviewable.io/reviews/openmoq/o-rly/34#-On1sUZM9whzYSYdv3ZO:-On1sUZM9whzYSYdv3ZP:b-i1u6i9) (no related file):*
Please format :smile: `scripts/format.sh`
___
*[a discussion](https://reviewable.io/reviews/openmoq/o-rly/34#-On1xQRz7s52KlKDhOIj:-On1xQRz7s52KlKDhOIk:b-8znrrp) (no related file):*
Do we want to have tests (even if AI generated) for stuff? 
___
*[`include/o_rly/admin/AdminServer.h` line 51 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1xoJ8ABdSS1aSJ5Vo:-On1xoJ8ABdSS1aSJ5Vp:bwlk289) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/include/o_rly/admin/AdminServer.h#L51)):*
> ```objc
> 
>   // Register a route. Must be called before start().
>   void addRoute(std::string method, std::string path, RouteHandler handler);
> ```

What happens if this is called after start?
I don't know if it's expected to be called from a single place, or if it's intended to be distributed when every piece registers its routes.

For single-place usage, doing it all in constructor may be better?
If distributed, then not calling this after start should be enforced. E.g., factory/builder class splitting up building and starting
___
*[`src/main.cpp` line 22 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1r4GEE2UG3E6KatlD:-On1r4GEE2UG3E6KatlE:b-vl6mbr) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/src/main.cpp#L22)):*
> ```cpp
> DEFINE_int32(max_cached_tracks, 100, "Maximum number of cached tracks (0 to disable caching)");
> DEFINE_int32(max_cached_groups_per_track, 3, "Maximum groups per track in cache");
> DEFINE_int32(admin_port, 9669, "HTTP admin port (0 to disable)");
> ```

I don't think "0 to disable" is wise; it seems nonobvious given that zero port has a special meaning.
I think explicit over implicit is the way to go with configuration. So something like `--disable-admin`. 
Also, it needs to be able to bind to a potentially different IP.

Generally I don't even like default ports and prefer explicit ones there, but that's just a personal preference. 

But this can be resolved later, not blocking on this. 
___
*[`src/main.cpp` line 75 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1rrgv-vACTeGph7OM:-On1rrgw4goU8X2IlPRA:b-12gpqa) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/src/main.cpp#L75)):*
> ```cpp
>     openmoq::o_rly::admin::registerBuiltinRoutes(adminServer);
>     if (!adminServer.start(static_cast<uint16_t>(FLAGS_admin_port))) {
>       XLOG(ERR) << "Failed to start admin server on port " << FLAGS_admin_port;
> ```

Why is this not a hard error that blocks the startup? Do you see a use case where its needed to start without admin if admin is configured? 
___
*[`src/admin/AdminServer.cpp` line 63 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On2-1BKCx1V0SiStGxg:-On2-1BKCx1V0SiStGxh:bxwtoa) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/src/admin/AdminServer.cpp#L63)):*
> ```cpp
>         return new AdminRequestHandler(route.handler);
>       }
>     }
> ```

I do expect that this matching will get more complicated as we get more routes and features. Basic linear exact match is fine now. 
___
*[`src/admin/AdminServer.cpp` line 68 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1vwD-6QicaDZTG9T0:-On1vwD-6QicaDZTG9T1:bbm49bh) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/src/admin/AdminServer.cpp#L68)):*
> ```cpp
>     return new AdminRequestHandler([](auto /*req*/, auto /*body*/, auto* downstream) {
>       proxygen::ResponseBuilder(downstream)
>           .status(404, "Not Found")
> ```

Nit: Is writing out the reason phrases a good idea/idiomatic? 
`proxygen::HTTPMessage::getDefaultReason`. I am kind of surprised that Response Builder doesn't have a method that takes only a status code and calls this internally. 

I know technically you can send any reason phrase you want, but that's not generally used nor supported. 
___
*[`src/admin/BuiltinRoutes.cpp` line 15 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1uQxf6IpWIEEWfart:-On1uQxf6IpWIEEWfaru:bnj1pz0) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/src/admin/BuiltinRoutes.cpp#L15)):*
> ```cpp
>         .status(200, "OK")
>         .header("Content-Type", "application/json")
>         .body(folly::IOBuf::copyBuffer("{\"service\":\"o-rly\",\"version\":\"0.1.0\"}\n"))
> ```

We should not drop the version as a random string all over the place, as we will 100% forget to update it. 
We already have 0.1.0 defined in `CMakeLists.txt:3`. 
But this is probably just a demo handler, so it doesn't matter?

src/admin/AdminServer.cpp
```diff
@@ -0,0 +1,112 @@
...
+AdminServer::AdminServer() = default;
+AdminServer::~AdminServer() = default;
+
+void AdminServer::addRoute(std::string method, std::string path, RouteHandler handler) {
+  routes_.push_back({std::move(method), std::move(path), std::move(handler)});
+}
+
+bool AdminServer::start(uint16_t port) {
+  proxygen::HTTPServerOptions options;
+  options.threads = 1;
```
gmarzot: define constant?

src/admin/BuiltinRoutes.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include <folly/io/IOBuf.h>
+#include <proxygen/httpserver/ResponseBuilder.h>
+
+#include <o_rly/admin/AdminServer.h>
+
+namespace openmoq::o_rly::admin {
+
+void registerBuiltinRoutes(AdminServer& server) {
+  server.addRoute("GET", "/info", [](auto /*req*/, auto /*body*/, auto* downstream) {
```
gmarzot, line 12: suggest some simple /health or /ping endpoint (maybe those are built in?)

src/main.cpp
```diff
@@ -18,6 +19,7 @@ DEFINE_string(
 );
 DEFINE_int32(max_cached_tracks, 100, "Maximum number of cached tracks (0 to disable caching)");
 DEFINE_int32(max_cached_groups_per_track, 3, "Maximum groups per track in cache");
+DEFINE_int32(admin_port, 9669, "HTTP admin port (0 to disable)");
```
gmarzot: suggest some HTTPS-like 9443?

gmarzot (approved):

src/main.cpp
afrind: :thumbsup:

src/admin/AdminServer.cpp
afrind: will do

src/admin/BuiltinRoutes.cpp
afrind: I think /health is a good endpoint to add but lets do as follow up?

afrind (commented):
> @afrind made 10 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/34)* status: all files reviewed, 10 unresolved discussions (waiting on afrind, gmarzot, and michalhosna).

___
*[a discussion](https://reviewable.io/reviews/openmoq/o-rly/34#-On1xQRz7s52KlKDhOIj:-On8FXqH2h0RJ4EjQO0i:b-qnyh8h) (no related file):*

Do we want to have tests (even if AI generated) for stuff?
</blockquote></details>

We don't really have a test framework for this since it's not unit testable.  You have to start the relay and e.g. curl it.  I can add a simple test script for it thoug.
___
*[a discussion](https://reviewable.io/reviews/openmoq/o-rly/34#-On1sUZM9whzYSYdv3ZO:-On8GbB8E4sss2OvS-gq:b-d7bhn0) (no related file):*

Please format :smile: `scripts/format.sh`
</blockquote></details>

will do
___
*[`include/o_rly/admin/AdminServer.h` line 51 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1xoJ8ABdSS1aSJ5Vo:-On8FlWX8MJH3oAL50PW:b-lsam13) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/include/o_rly/admin/AdminServer.h#L51)):*
> if it's intended to be distributed when every piece registers its routes.

I see it more like this.  Different components can register during their init, maybe?  I will add a CHECK that start hasn't been called.
___
*[`src/main.cpp` line 22 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1r4GEE2UG3E6KatlD:-On8FzJ96sNvZg13P5Sw:b-16s5zz) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/src/main.cpp#L22)):*

I don't think "0 to disable" is wise; it seems nonobvious given that zero port has a special meaning.
I think explicit over implicit is the way to go with configuration. So something like `--disable-admin`. 
Also, it needs to be able to bind to a potentially different IP.

Generally I don't even like default ports and prefer explicit ones there, but that's just a personal preference. 

But this can be resolved later, not blocking on this.
</blockquote></details>

Ok yes.  I'll just remove this option to disable entirely.  It's all moving to config anyways.
___
*[`src/main.cpp` line 75 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1rrgv-vACTeGph7OM:-On8GD4I-3WfECGRAv7Z:bfwm7e5) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/src/main.cpp#L75)):*

Why is this not a hard error that blocks the startup? Do you see a use case where its needed to start without admin if admin is configured?
</blockquote></details>

I can make it a hard error.
___
*[`src/admin/AdminServer.cpp` line 68 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1vwD-6QicaDZTG9T0:-On8GKvf64AFUBqEe2tI:b-maf1rt) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/src/admin/AdminServer.cpp#L68)):*

Nit: Is writing out the reason phrases a good idea/idiomatic? 
`proxygen::HTTPMessage::getDefaultReason`. I am kind of surprised that Response Builder doesn't have a method that takes only a status code and calls this internally. 

I know technically you can send any reason phrase you want, but that's not generally used nor supported.
</blockquote></details>

1) good point
2) H2 doesn't even have reason phrase lol
___
*[`src/admin/BuiltinRoutes.cpp` line 15 at r1](https://reviewable.io/reviews/openmoq/o-rly/34#-On1uQxf6IpWIEEWfart:-On8GYZq1d3Hw-mH-t8f:b-rn0cad) ([raw file](https://github.com/openmoq/o-rly/blob/cec5b5bf2a965f42b589e5e83f9e8727c6006295/src/admin/BuiltinRoutes.cpp#L15)):*

We should not drop the version as a random string all over the place, as we will 100% forget to update it. 
We already have 0.1.0 defined in `CMakeLists.txt:3`. 
But this is probably just a demo handler, so it doesn't matter?
</blockquote></details>

This is just a demo, will see if we can insert the version from the cmake constant.

michalhosna (approved):
> @michalhosna reviewed 7 files and all commit messages, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/34)* status: all files reviewed, 4 unresolved discussions (waiting on afrind and gmarzot).

src/admin/AdminServer.cpp
afrind: Done.

afrind (commented):
> @afrind made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/34)* status: 7 of 8 files reviewed, 4 unresolved discussions (waiting on afrind, gmarzot, and michalhosna).

___
*[a discussion](https://reviewable.io/reviews/openmoq/o-rly/34#-On1sUZM9whzYSYdv3ZO:-OnIJsC90bdN-kWrvDuB:bipy5q6) (no related file):*

will do
</blockquote></details>

main is formatted.  Not reformatting files I didn't touch.

michalhosna (approved):
> @michalhosna reviewed 2 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/34)* status: all files reviewed, 3 unresolved discussions (waiting on gmarzot).

src/main.cpp
afrind: Hm, maybe a bikeshed, since right now it's cleartext (9080?).  Suggest we defer.

afrind (commented):
> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/34)* status: all files reviewed, 3 unresolved discussions (waiting on gmarzot).

gmarzot (commented):
> @gmarzot reviewed 8 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/34)* status: all files reviewed, 2 unresolved discussions (waiting on afrind).

gmarzot (approved):
> @gmarzot resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/34)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #36, opened 2026-03-06
config: use macros for schema definition

michalhosna:
I asked Claude to try to redefine the config using macros. 

It's nice that we cannot, accidentally, have different defaults in documentation and during runtime.
But it's adding a level of indirection. Adds a macro we have to maintain, and I don't think it really saves that much code. 

I am a proponent of explicit over implicit when it comes to configuration. I don't expect many defaults in the config down the line. Not only that, but I would even throw out the default port and default IP. 

For fields without defaults, this doesn't really add anything, I think.

I would personally not use this, but I am not opposed to using this. If @afrind thinks it's better, let's go with it.

@afrind your call.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/36)


michalhosna:
I dropped the defaults, so this now is really very pointless.

afrind (changes requested):
Well the purpose of the CONFIG_OPT was to supply the default and define the non-optional accessor, so this is incomplete in my view.  But we talked about this and if you prefer separating the parsed config and used config that has different tradeoffs and I can live with it.

> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/36#-:-On8ODpd29VuBYB8dNV9:b-jltsfw)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna).

PR #40, opened 2026-03-08
macOS CI, notifications, rolling snapshot release

gmarzot:
## Summary

- **Add macOS to CI and publish** — `macos-15` runner in both matrices
- **Rename workflows** — `build` → `verify`, `publish release` → `publish artifacts`
- **Fix notify-main-failure trigger** — `"build"` → `"verify"` (same class of bug as moxygen #56)
- **Standardize notifications** — bold `*org/repo*`, backtick branch/hash, verbose email
- **Update `setup-deps-tarball.sh`** — download from `snapshot-latest` instead of `build-<sha>` tag
- **Advance moxygen submodule** to `b4e4abd9` (includes #56 changes)

## Changes

| File | What |
|------|------|
| `ci.yml` | `name: verify`, add macOS matrix, `nproc` → `getconf _NPROCESSORS_ONLN` |
| `publish.yml` | `name: publish artifacts`, macOS matrix, Docker conditioned on Linux, snapshot-latest, standardized notifications |
| `notify-main-failure.yml` | Trigger `"verify"`, 7-char hashes, bold/code Slack format |
| `setup-deps-tarball.sh` | Download from `snapshot-latest` instead of `build-<sha>` |
| `deps/moxygen` | Submodule → `b4e4abd9` |

## Depends on

- [x] openmoq/moxygen#56 (merged)
- [ ] openmoq/moxygen#57 (Slack formatting — cosmetic, not blocking)

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/40)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,4 +61,5 @@ jobs:
           name: "test (${{ matrix.name }})"
           path: ${{ matrix.build_dir }}/test-results.xml
           reporter: java-junit
+          fail-on-empty: 'false'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: build
+name: verify
 
 on:
   pull_request:
@@ -19,11 +19,17 @@ jobs:
           - name: linux
             preset: default
             build_dir: build
+            runner: ubuntu-22.04
+          - name: macos
+            preset: default
+            build_dir: build
+            runner: macos-15
           - name: asan debug
             preset: san
             build_dir: build-san
+            runner: ubuntu-22.04
     name: ${{ matrix.name }}
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.runner }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -41,7 +47,7 @@ jobs:
         run: bash scripts/configure.sh ${{ matrix.build_dir }} ${{ matrix.preset }}
 
       - name: Build
-        run: cmake --build ${{ matrix.build_dir }} -j$(nproc)
+        run: cmake --build ${{ matrix.build_dir }} -j$(getconf _NPROCESSORS_ONLN)
 
       - name: Test
         env:
--- a/.github/workflows/notify-main-failure.yml
+++ b/.github/workflows/notify-main-failure.yml
@@ -25,7 +25,7 @@ jobs:
           SHORT="${{ github.event.workflow_run.head_sha }}"
           SHORT="${SHORT:0:7}"
           RUN_URL="${{ github.event.workflow_run.html_url }}"
-          TEXT=":x: *${{ github.repository }}* \`main\` \`${SHORT}\` — build failed: Run <${RUN_URL}|#${{ github.event.workflow_run.run_number }}>"
+          TEXT=":x: *${{ github.repository }}* \`main\` \`${SHORT}\` — verify failed: <${RUN_URL}|#${{ github.event.workflow_run.run_number }}>"
           curl -s -X POST "$SLACK_WEBHOOK_URL" \
             -H "Content-Type: application/json" \
             --data "{\"text\": \"${TEXT}\"}"
@@ -40,8 +40,8 @@ jobs:
           SHORT="${{ github.event.workflow_run.head_sha }}"
           SHORT="${SHORT:0:7}"
           RUN_URL="${{ github.event.workflow_run.html_url }}"
-          SUBJECT="[o-rly] build failed on main ${SHORT}"
-          BODY="Build failed on main.\n\nCommit: ${{ github.event.workflow_run.head_sha }}\nRun: ${RUN_URL}"
+          SUBJECT="[o-rly] verify failed on main ${SHORT}"
+          BODY="Verify failed on main.\n\nCommit: ${{ github.event.workflow_run.head_sha }}\nRun: ${RUN_URL}"
           aws ses send-email \
             --from "noreply@ci.openmoq.org" \
             --destination '{"ToAddresses":["github-notifications@openmoq.org"]}' \
--- a/.github/workflows/notify-main-failure.yml
+++ b/.github/workflows/notify-main-failure.yml
@@ -1,11 +1,11 @@
 name: notify main failure
 
-# Fires only when the build workflow completes on main (push events).
+# Fires only when the verify workflow completes on main (push events).
 # PR branch builds never match the branches: [main] filter.
 
 on:
   workflow_run:
-    workflows: ["build"]
+    workflows: ["verify"]
     types: [completed]
     branches: [main]
 
@@ -23,9 +23,9 @@ jobs:
           SLACK_WEBHOOK_URL: ${{ secrets.OMOQ_SLACK_WEBHOOK_URL }}
         run: |
           SHORT="${{ github.event.workflow_run.head_sha }}"
-          SHORT="${SHORT:0:12}"
+          SHORT="${SHORT:0:7}"
           RUN_URL="${{ github.event.workflow_run.html_url }}"
-          TEXT=":x: *${{ github.repository }}* \`main\` CI failed — <${RUN_URL}|run>"
+          TEXT=":x: *${{ github.repository }}* \`main\` \`${SHORT}\` CI failed — run: <${RUN_URL}|view>"
           curl -s -X POST "$SLACK_WEBHOOK_URL" \
             -H "Content-Type: application/json" \
             --data "{\"text\": \"${TEXT}\"}"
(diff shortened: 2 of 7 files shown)


michalhosna (changes requested):
> @michalhosna reviewed 1 file and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/40)* status: 1 of 7 files reviewed, 5 unresolved discussions (waiting on gmarzot).

___
*[`.github/workflows/publish.yml` line 24 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMr3hN1ac6_m286AvS:-OnMr3hN1ac6_m286AvT:bat1swp) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/.github/workflows/publish.yml#L24)):*
> ```YAML
>             runner: ubuntu-22.04
>           - name: macos-15-arm64
>             runner: macos-15
> ```

Given that macOS is not deliverable, I don't think it should fail the whole CI if it doesn't pass. 

I don't really love spending time on macOS at all at this point. I won't be dying on this hill. 
But I don't think the entire CI should fail just because macOS fails. It should be a warning at most. I am not sure what the GH actions options are there. 


 
___
*[`.github/workflows/publish.yml` line 124 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMp5hK2sSN4AeUi-P_:-OnMp5hK2sSN4AeUi-Pa:b-64utun) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/.github/workflows/publish.yml#L124)):*
> ```YAML
>           git tag -d "$TAG" 2>/dev/null || true
>           git push origin ":refs/tags/$TAG" 2>/dev/null || true
> ```

Isn't this racy? If two pushes to the main happen right after each other, one may overwrite the latest of the other one. 

I am not sure what the GH Actions defaults are. Looking around docs, something like. 
```
  concurrency:
    group: publish
    cancel-in-progress: true
```
Should solve the problem, as a newer publication always cancels the old one. 

Or maybe limiting concurrency to one, so it must run sequentially. 
___
*[`docker/Dockerfile` line 1 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMqSuu1yCGtfYV33qu:-OnMqSuu1yCGtfYV33qv:b-vuym02) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/docker/Dockerfile#L1)):*
> ```Dockerfile
> FROM ubuntu:22.04
> ```

I don't understand why the switch to ubuntu? 
___
*[`scripts/setup-deps-tarball.sh` line 98 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMlWOlEyH0bkWGZKL9:-OnMlWOlEyH0bkWGZKLA:bhefymr) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/scripts/setup-deps-tarball.sh#L98)):*
> ```Shell
>         --pattern "$TARBALL" \
>         --dir "$DOWNLOAD_DIR"
>     echo "  Warning: using snapshot-latest (may not match submodule SHA)"
> ```

What's the thinking behind using snapshot-latest if it may be a different code? Would it be better to just fail hard? 

I don't know if this is oriented strictly for CI use or if it is ever expected this to be run at developer machines? 

In CI, I don't want CI to pass if I mangled something with submodules. I would prefer a hard fail, or at least something more visible than a plain echo line inside a log that I don't think anyone opens if something passes. 
___
*[`.github/workflows/verify.yml` line 71 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMlyRmCWNjwqO4MGoh:-OnMlyRmCWNjwqO4MGoi:b-w7s23r) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/.github/workflows/verify.yml#L71)):*
> ```YAML
>           path: ${{ matrix.build_dir }}/test-results.xml
>           reporter: java-junit
>           fail-on-empty: 'false'
> ```

This now applies to the whole matrix. 
If I understand this correctly, if anybody for any reason breaks the running of tests in CI, this will now siletly accept this?

If it's only because of macOS, we should split macos to a separate job all together.

gmarzot:
*[`.github/workflows/publish.yml` line 124 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMp5hK2sSN4AeUi-P_:-OnNsNZl6NXvV4SfYzZ1:bgv0qxv) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/.github/workflows/publish.yml#L124)):*

Isn't this racy? If two pushes to the main happen right after each other, one may overwrite the latest of the other one. 

I am not sure what the GH Actions defaults are. Looking around docs, something like. 
```
  concurrency:
    group: publish
    cancel-in-progress: true
```
Should solve the problem, as a newer publication always cancels the old one. 

Or maybe limiting concurrency to one, so it must run sequentially.
</blockquote></details>

good catch - fixing

gmarzot:
*[`.github/workflows/verify.yml` line 71 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMlyRmCWNjwqO4MGoh:-OnNy8ez5aZZO2f-sHIe:b-8fe1lu) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/.github/workflows/verify.yml#L71)):*

This now applies to the whole matrix. 
If I understand this correctly, if anybody for any reason breaks the running of tests in CI, this will now siletly accept this?

If it's only because of macOS, we should split macos to a separate job all together.
</blockquote></details>

improved.
 `fail-on-empty: ${{ job.status == 'success' && 'true' || 'false' }}` — catches missing XML when build succeeded, tolerates it when build failed.

gmarzot:
*[`scripts/setup-deps-tarball.sh` line 98 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMlWOlEyH0bkWGZKL9:-OnNyJ6FEDSCoPa08qMu:b-l2vwoi) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/scripts/setup-deps-tarball.sh#L98)):*

What's the thinking behind using snapshot-latest if it may be a different code? Would it be better to just fail hard? 

I don't know if this is oriented strictly for CI use or if it is ever expected this to be run at developer machines? 

In CI, I don't want CI to pass if I mangled something with submodules. I would prefer a hard fail, or at least something more visible than a plain echo line inside a log that I don't think anyone opens if something passes.
</blockquote></details>

fixed. artifacts used purely identified by submodule hash. otherwise fail - no fall back (claude's fault !)

gmarzot:
*[`docker/Dockerfile` line 1 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMqSuu1yCGtfYV33qu:-OnO03pyDMrvoFXjXCr2:b-fmrk0j) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/docker/Dockerfile#L1)):*

I don't understand why the switch to ubuntu?
</blockquote></details>

that was interim while trying to nail down the static linking in moxygen pr #58 . back to bookworm-slim now and the static linking mission success . additional deps down to 2 from 10

gmarzot:
*[`.github/workflows/publish.yml` line 24 at r3](https://reviewable.io/reviews/openmoq/o-rly/40#-OnMr3hN1ac6_m286AvS:-OnNsDcYAPuN_dznqdbb:b7ig0b6) ([raw file](https://github.com/openmoq/o-rly/blob/5d4a6904fa955fc921ac30efe8c82dd1b21168fd/.github/workflows/publish.yml#L24)):*

Given that macOS is not deliverable, I don't think it should fail the whole CI if it doesn't pass. 

I don't really love spending time on macOS at all at this point. I won't be dying on this hill. 
But I don't think the entire CI should fail just because macOS fails. It should be a warning at most. I am not sure what the GH actions options are there.
</blockquote></details>

i will defer on it . pulled from relay ci for now. I think it will be a thing later but you are right better to land the rest of this now

michalhosna (approved):
:lgtm:

> @michalhosna partially reviewed 6 files, made 1 comment, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/40#-:-OnO3WH_2Z-a8gE8eOA3:bnfp4nl)* status: 6 of 7 files reviewed, all discussions resolved (waiting on gmarzot).

PR #42, opened 2026-03-09
CI: Add format check

michalhosna:
Add --check flag to scripts/format.sh rather than calling clang-format
directly in ci.yml, so the set of matched files stays consistent between
local formatting and CI verification.

Replace rg with find in the script to avoid requiring ripgrep in CI.

Rename workflow from "build" to "ci" to match the filename and reflect
that it now covers more than building.

The format job uses ubuntu-24.04 (unlike the build jobs) to get a
reasonably modern clang-format, even though we target older compilers
for building.
  

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/42)

PR #43, opened 2026-03-09
Run format.sh

afrind:
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/43)


michalhosna:
I think we are maybe running into clang version differences given that #42 passes. 
@afrind What clang format version do you have? What version do we want?

afrind:
openmoq-dev-1:~/develop/o-rly afrind# clang-format --version
clang-format version 21.1.8 (CentOS 21.1.8-1.el10)

michalhosna:
I would suggest closing in favor of #42, where we should handle the versioning once and for all.

PR #49, opened 2026-03-11
Add Concrete Implementations for Stats Collection

akash-a-n:
- Added MoQStats Collector

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49)


afrind (commented):
> @afrind reviewed 1 file and all commit messages, and made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49)* status: 1 of 10 files reviewed, 9 unresolved discussions (waiting on akash-a-n).

___
*[`include/o_rly/stats/MoQStatsCollector.h` line 25 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTY5P--K6Qapttiene:-OnTY5P--K6Qapttienf:bnd29vh) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/include/o_rly/stats/MoQStatsCollector.h#L25)):*
> ```objc
> // ---------------------------------------------------------------------------
> 
> template <size_t N> struct BoundedHistogram {
> ```

Hmm it didn't like the one in folly, or there wasn't one?
___
*[`src/ORelayServer.cpp` line 64 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYi2CBXq7zRKB01uD:-OnTYi2CBXq7zRKB01uE:b-eolsc) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/ORelayServer.cpp#L64)):*
> ```cpp
> 
>   if (statsRegistry_) {
>     // Create a collector for this session and register it with the registry.
> ```

I don't think we want a stats collector per session, I think we want it per executor (so for now, just one)
___
*[`src/admin/MetricsHandler.cpp` line 28 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYzEv-DzgnzDzaQ8D:-OnTYzEv-DzgnzDzaQ8E:b-jal7li) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L28)):*
> ```cpp
>         // response asynchronously. This avoids blocking the admin thread.
>         registry->aggregateAsync()
>             .scheduleOn(relayExecutor.get())
> ```

The register may have more than one executor in the future, so I think it should handle this detail internally.
___
*[`src/admin/MetricsHandler.cpp` line 32 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZAP-4eOlZr8EW1Qs:-OnTZAP-4eOlZr8EW1Qt:b-r1cqbo) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L32)):*
> ```cpp
>               if (result.hasException()) {
>                 try {
>                   result.throwUnlessValue();
> ```

Send a 500 http response instead of throwing?
___
*[`src/admin/MetricsHandler.cpp` line 44 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZIbw-bNDk4HX0w1s:-OnTZIbw-bNDk4HX0w1t:b2zzl2s) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L44)):*
> ```cpp
> 
>               const auto& snapshot = result.value();
>               auto body = stats::StatsSnapshot::formatPrometheus(snapshot);
> ```

Should we have formatPrometheus serialize to an IOBuf chain instead of a string?  Then we can avoid the copyBuffer below
___
*[`src/stats/MoQStatsCollector.cpp` line 11 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZYJs1zSg8TXkAwPm:-OnTZYJs1zSg8TXkAwPn:b-wjr782) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/MoQStatsCollector.cpp#L11)):*
> ```cpp
>     std::shared_ptr<StatsRegistry> registry
> )
>     : owningExecutor_(std::move(owningExecutor)), registry_(registry) {}
> ```

Do you want the ctor to regsiter (since dtor unregisters?)

Also, register should take the exec
___
*[`src/stats/MoQStatsCollector.cpp` line 135 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZndw9ZSSpifaGgkd:-OnTZndw9ZSSpifaGgke:bn6ccs8) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/MoQStatsCollector.cpp#L135)):*
> ```cpp
> void MoQStatsCollector::onPublishSuccess() {
>   ++moqPublishSuccess_;
>   ++moqActivePublishers_;
> ```

Not sure, this may be "active publish subscriptions" rather than publishers.
___
*[`src/stats/StatsRegistry.cpp` line 112 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnT_BXz5DCtA09KRy6h:-OnT_BXz5DCtA09KRy6i:b2ubif0) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/StatsRegistry.cpp#L112)):*
> ```cpp
> 
> void StatsRegistry::onNewSession() {
>   std::lock_guard lock(mu_);
> ```

We should avoid locks for stats.  I think this metric can be captured in the collector, which is thread-local anyways?
___
*[`src/stats/StatsRegistry.cpp` line 134 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnT_6ZL2BGCfNYYfgH5:-OnT_6ZL2BGCfNYYfgH6:b-l2vklu) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/StatsRegistry.cpp#L134)):*
> ```cpp
>   }
> 
>   // --- Aggregate collectors ---
> ```

Yes, do this :)

akash-a-n (commented):
> @akash-a-n made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49)* status: 1 of 10 files reviewed, 9 unresolved discussions (waiting on afrind).

___
*[`include/o_rly/stats/MoQStatsCollector.h` line 25 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTY5P--K6Qapttiene:-OnWF8Hb6AE0TfZGE56V:bf7sa02) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/include/o_rly/stats/MoQStatsCollector.h#L25)):*

Hmm it didn't like the one in folly, or there wasn't one?
</blockquote></details>

Folly doesn't have a native Histogram for varied bucket sizes, hence it suggested this.
___
*[`src/ORelayServer.cpp` line 64 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYi2CBXq7zRKB01uD:-OnXa9G203_arpZySEuu:bfjd4eo) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/ORelayServer.cpp#L64)):*

I don't think we want a stats collector per session, I think we want it per executor (so for now, just one)
</blockquote></details>

Done.
I have added a concurrent hashmap to store each collector corresponding to an executor. Overkill for now? Did you mean to have just one collector instead?
___
*[`src/admin/MetricsHandler.cpp` line 28 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYzEv-DzgnzDzaQ8D:-OnXJ97c4cp-rAtn8lBu:b4mse3v) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L28)):*

The register may have more than one executor in the future, so I think it should handle this detail internally.
</blockquote></details>

Aggregation for each collector happens on its own executor. This is Aggregation across executors. So I have updated it to be a blocking wait for now. 
Does proxygen server offer any async mechanisms? Or add more threads? Do you have other ideas about this?
___
*[`src/admin/MetricsHandler.cpp` line 32 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZAP-4eOlZr8EW1Qs:-OnWFo877pay5RN3Hc9l:b-qwo1yr) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L32)):*

Send a 500 http response instead of throwing?
</blockquote></details>

Removed the try/catch indirection
___
*[`src/admin/MetricsHandler.cpp` line 44 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZIbw-bNDk4HX0w1s:-OnXZzeHDp6BwnqgL7GA:bb6un1j) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L44)):*

Should we have formatPrometheus serialize to an IOBuf chain instead of a string?  Then we can avoid the copyBuffer below
</blockquote></details>

Done
___
*[`src/stats/MoQStatsCollector.cpp` line 11 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZYJs1zSg8TXkAwPm:-OnXYXgPB1QcrJfIOcLB:b-8ommse) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/MoQStatsCollector.cpp#L11)):*
moved registering to a create function that registers after calling the private constructor. 

> Also, register should take the exec
I don't understand the rationale, We have a single global register and a per-executor collector, so every collector gets an executor. Why would we pass along the exec to the register?
___
*[`src/stats/MoQStatsCollector.cpp` line 135 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZndw9ZSSpifaGgkd:-OnXjCNx22ZPaM1m4n--:bhvvl13) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/MoQStatsCollector.cpp#L135)):*

Not sure, this may be "active publish subscriptions" rather than publishers.
</blockquote></details>

Will look into each of the metrics thoroughly 
___
*[`src/stats/StatsRegistry.cpp` line 112 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnT_BXz5DCtA09KRy6h:-OnXjJlc075YjjqjLItG:b-2gvn11) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/StatsRegistry.cpp#L112)):*

We should avoid locks for stats.  I think this metric can be captured in the collector, which is thread-local anyways?
</blockquote></details>

Yes. Now that collectors are per-executor, moved number of session metric into the collector.
___
*[`src/stats/StatsRegistry.cpp` line 134 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnT_6ZL2BGCfNYYfgH5:-OnXVrPp9dQcta2szbTd:bb7f2mf) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/StatsRegistry.cpp#L134)):*

Yes, do this :)
</blockquote></details>

done

afrind (commented):
Ok, coming along.  It would probably be good to add some kind of tests for this.  The admin server tests just spawn the relay and curl the endpoint.  We could do something similar for now, though I guess without a moq test client all the counters will be zero.  Still, it's something

> @afrind reviewed 9 files and all commit messages, made 10 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49#-:-OnZK3W1-HLPJoTpK_52:b9841mo)* status: all files reviewed, 8 unresolved discussions (waiting on akash-a-n).

___
*[`include/o_rly/stats/MoQStatsCollector.h` line 25 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTY5P--K6Qapttiene:-OnZ5JpOAkjBkkaA3sZq:bqfw0zm) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/include/o_rly/stats/MoQStatsCollector.h#L25)):*

Folly doesn't have a native Histogram for varied bucket sizes, hence it suggested this.
</blockquote></details>

can we move this to it's own file with unit test, possibly it's own commit?
___
*[`src/ORelayServer.cpp` line 64 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYi2CBXq7zRKB01uD:-OnZ6xII4U3hjrK-0pS4:bxb0ndk) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/ORelayServer.cpp#L64)):*

Done.
I have added a concurrent hashmap to store each collector corresponding to an executor. Overkill for now? Did you mean to have just one collector instead?
</blockquote></details>

This isn't quite what I want.  I was thinking something like EventBaseLocal, except we can't guarantee that exec is an event base.  For now it probably _is_ overkill since the relay isn't mt safe anyways.  So you can just hold a single collector.
___
*[`src/admin/MetricsHandler.cpp` line 28 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYzEv-DzgnzDzaQ8D:-OnZ96Q4B2rBasdetuWx:bd2lm5c) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L28)):*

Aggregation for each collector happens on its own executor. This is Aggregation across executors. So I have updated it to be a blocking wait for now. 
Does proxygen server offer any async mechanisms? Or add more threads? Do you have other ideas about this?
</blockquote></details>

You don't want blocking wait.  I thought your previous code had it right - launch the coroutine using .start() and you can follow that with a lambda which runs on completion.  We do need to be careful with cancellation.  There's folly::coro::AsyncScope but when I played with it years ago it didn't do quite what I wanted.  Maybe see if it's apt, but to do manually I would:

1. create a folly::CancellationSource in the request handler (that's getting the onHeaders/onBody etc callbacks)
2. launch the coroutine using co_withCancellation and cancellationSource_.getToken()
3. if requestError happens, invoke cancellationSource_.requestCancellation()

So if the request is aborted by the client before the response is produced, cancellation should propagate to collectAll and it will yield an exception, which hopefully we translate to no-op.  You can check the token explicitly if needed too (pass in with getToken, then if (token.isCancellationRequested()).
___
*[`src/stats/MoQStatsCollector.cpp` line 11 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZYJs1zSg8TXkAwPm:-OnZBjov6tyOqE5s2PKD:bbj2vbt) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/MoQStatsCollector.cpp#L11)):*

moved registering to a create function that registers after calling the private constructor. 

> Also, register should take the exec
I don't understand the rationale, We have a single global register and a per-executor collector, so every collector gets an executor. Why would we pass along the exec to the register?
</blockquote></details>

aggregateAsync needs to use co_withExecutor so that the snapshot task runs in the owningExecutor.
___
*[`src/ORelayServer.cpp` line 68 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZ77nM7MMSgDjdKEyC:-OnZ77nM7MMSgDjdKEyD:b-84ie3n) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/ORelayServer.cpp#L68)):*
> ```cpp
>     auto it = executorCollectors_.find(execKey);
>     if (it == executorCollectors_.end()) {
>       auto collector = stats::MoQStatsCollector::create_moq_stats_collector(
> ```

camelCase createMoQStatsCollector?
___
*[`src/ORelayServer.cpp` line 69 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZ8KNI3FSpvInyHfPG:-OnZ8KNI3FSpvInyHfPH:blpd3d6) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/ORelayServer.cpp#L69)):*
> ```cpp
>     if (it == executorCollectors_.end()) {
>       auto collector = stats::MoQStatsCollector::create_moq_stats_collector(
>           folly::getKeepAliveToken(*clientSession->getExecutor()),
> ```

Unsure if we want a keepalive here or not.  The keepalive will keep the alive, but I don't think stats requires that?  I know the raw ptr is a little sketchy though.  Maybe we hold keepalive but there's a shutdown method that releases them?  I'd ask AI.

But if we're just holding one for now, then it's moot anyways.
___
*[`src/admin/MetricsHandler.cpp` line 24 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZATVj5oHf4vRc9tcD:-OnZATVj5oHf4vRc9tcE:beulvqf) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/admin/MetricsHandler.cpp#L24)):*
> ```cpp
>       [registry = std::move(registry)](auto /*req*/, auto /*body*/, auto* downstream) {
>         try {
>           const auto& snapshot = folly::coro::blockingWait(registry->aggregateAsync());
> ```

holding this in a reference feels sketchy (a reference to a temp return value?), did auto snapshot not work?
___
*[`src/stats/MoQStatsCollector.cpp` line 13 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZAnwBA4Y3PR0PTL0S:-OnZAnwBA4Y3PR0PTL0T:b-sb76oo) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/stats/MoQStatsCollector.cpp#L13)):*
> ```cpp
> ) {
>   auto collector =
>       std::shared_ptr<MoQStatsCollector>(new MoQStatsCollector(std::move(owningExecutor)));
> ```

use std::make_shared<MoQStatsCollector>) it's less verbose and allocates the reference block contiguous with the object memory
___
*[`src/stats/StatsRegistry.cpp` line 125 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZC5t4BWrS6bN-KSyW:-OnZC5t4BWrS6bN-KSyX:b68241z) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/stats/StatsRegistry.cpp#L125)):*
> ```cpp
>   StatsSnapshot combined;
>   {
>     std::lock_guard lock(mu_);
> ```

We want to avoid this lock too.  It exists because we lazily create collectors so in theory someone could add a collector during collection.  Let's require pre-registration instead, register a single one, the add a ::lock() API that will cause register to XLOG(FATAL) if someone later calls register

akash-a-n (commented):
Will add some tests.

> @akash-a-n made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49#-:-OnbvDoQ7NKQs4Ey2i27:b-haoamn)* status: 2 of 15 files reviewed, 8 unresolved discussions (waiting on afrind).

___
*[`include/o_rly/stats/MoQStatsCollector.h` line 25 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTY5P--K6Qapttiene:-OnaYeuT8XrD-0DGs6EK:b-896fix) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/include/o_rly/stats/MoQStatsCollector.h#L25)):*

can we move this to it's own file with unit test, possibly it's own commit?
</blockquote></details>

Done.
___
*[`src/ORelayServer.cpp` line 64 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYi2CBXq7zRKB01uD:-OnaYi541G8cfdPmF8JY:b-ujbara) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/ORelayServer.cpp#L64)):*

This isn't quite what I want.  I was thinking something like EventBaseLocal, except we can't guarantee that exec is an event base.  For now it probably _is_ overkill since the relay isn't mt safe anyways.  So you can just hold a single collector.
</blockquote></details>

Done. Simplified to a single collector
___
*[`src/ORelayServer.cpp` line 69 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZ8KNI3FSpvInyHfPG:-Onbpl4T2ZStuxeYkApW:bwwpuiq) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/ORelayServer.cpp#L69)):*

Unsure if we want a keepalive here or not.  The keepalive will keep the alive, but I don't think stats requires that?  I know the raw ptr is a little sketchy though.  Maybe we hold keepalive but there's a shutdown method that releases them?  I'd ask AI.

But if we're just holding one for now, then it's moot anyways.
</blockquote></details>

We need the executor for aggregation. Now that the collector_ is bound by the ORelayServer, the keepAlive reference would get destroyed with the server. 
Do you think we need an explicit shutdown tied to MoQServer::stop?
___
*[`src/admin/MetricsHandler.cpp` line 28 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYzEv-DzgnzDzaQ8D:-Onbbb_CE6Jetphya-68:bwel0g5) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L28)):*

You don't want blocking wait.  I thought your previous code had it right - launch the coroutine using .start() and you can follow that with a lambda which runs on completion.  We do need to be careful with cancellation.  There's folly::coro::AsyncScope but when I played with it years ago it didn't do quite what I wanted.  Maybe see if it's apt, but to do manually I would:

1. create a folly::CancellationSource in the request handler (that's getting the onHeaders/onBody etc callbacks)
2. launch the coroutine using co_withCancellation and cancellationSource_.getToken()
3. if requestError happens, invoke cancellationSource_.requestCancellation()

So if the request is aborted by the client before the response is produced, cancellation should propagate to collectAll and it will yield an exception, which hopefully we translate to no-op.  You can check the token explicitly if needed too (pass in with getToken, then if (token.isCancellationRequested()).
</blockquote></details>

Implemented this. Learning a lot about folly/async. I still have one question. 
Which executor should co_withCancellation use? In the current implementation, claude insisted on using `folly::getGlobalCPUExecutor(). `
The other option being proxygen's evb. -- `downstream->getTransaction()->getEventBase()`. Somehow I feel I should use proxygen's evb, But I don't have enough experience to disagree with the AI :'D
What do you think?
___
*[`src/admin/MetricsHandler.cpp` line 24 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZATVj5oHf4vRc9tcD:-OnbdAxaAWDjLheS3Gyb:b-hfpv0l) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/admin/MetricsHandler.cpp#L24)):*

holding this in a reference feels sketchy (a reference to a temp return value?), did auto snapshot not work?
</blockquote></details>

removed
___
*[`src/stats/MoQStatsCollector.cpp` line 11 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTZYJs1zSg8TXkAwPm:-OnbdJWw9NsU4yFJcNxe:b6dc9ym) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/stats/MoQStatsCollector.cpp#L11)):*

aggregateAsync needs to use co_withExecutor so that the snapshot task runs in the owningExecutor.
</blockquote></details>

yes you are right. Missed implementing it somehow. So registry has a reference to collection and collection currently owns the executor reference. So we can use `co_withExecutor(collector->exec)`
___
*[`src/stats/StatsRegistry.cpp` line 125 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZC5t4BWrS6bN-KSyW:-Onbv9b6CmPzgDf1JlTU:b-896fix) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/stats/StatsRegistry.cpp#L125)):*

We want to avoid this lock too.  It exists because we lazily create collectors so in theory someone could add a collector during collection.  Let's require pre-registration instead, register a single one, the add a ::lock() API that will cause register to XLOG(FATAL) if someone later calls register
</blockquote></details>

Done.

akash-a-n (commented):
I have added a separate script for this endpoint for now. I see some scope for improvement. We could spawn the server separately and combine the test scripts. WDYT?

> @akash-a-n made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49#-:-OnrCxjj9fQxut4WPI3w:b1xsis5)* status: 2 of 15 files reviewed, 8 unresolved discussions (waiting on afrind).

___
*[`src/admin/MetricsHandler.cpp` line 28 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYzEv-DzgnzDzaQ8D:-OnrCZj5-RQLbCGQ5A94:b-fxm4qc) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L28)):*

Implemented this. Learning a lot about folly/async. I still have one question. 
Which executor should co_withCancellation use? In the current implementation, claude insisted on using `folly::getGlobalCPUExecutor(). `
The other option being proxygen's evb. -- `downstream->getTransaction()->getEventBase()`. Somehow I feel I should use proxygen's evb, But I don't have enough experience to disagree with the AI :'D
What do you think?
</blockquote></details>

Actually, I decided to use `downstream->getTransaction()->getEventBase()` as waking up a OS thread for this was sub-optimal.

akash-a-n (commented):
Some of the changes after the last review:
- Split metrics into pub-sub
- Added a test script
- Each snapshot is generated in the owning executor
- Removed blocking wait. Using evb thread to schedule it with cancel token

> @akash-a-n made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49#-:-OoAQNX7C8KqDf0N4euX:bahb21w)* status: 0 of 16 files reviewed, 8 unresolved discussions (waiting on afrind).

akash-a-n (commented):
> @akash-a-n made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49)* status: 0 of 16 files reviewed, 8 unresolved discussions (waiting on afrind).

___
*[`src/admin/MetricsHandler.cpp` line 28 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYzEv-DzgnzDzaQ8D:-OoARLRq-15Oh3ue3kYn:b-b3xar1) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L28)):*

Actually, I decided to use `downstream->getTransaction()->getEventBase()` as waking up a OS thread for this was sub-optimal.
</blockquote></details>

modified to use `folly::EventBaseManager::get()->getEventBase();` for simplicity

afrind (approved):
> @afrind reviewed 16 files and all commit messages, made 10 comments, and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49)* status: all files reviewed, 1 unresolved discussion (waiting on akash-a-n).

___
*[`src/ORelayServer.cpp` line 69 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZ8KNI3FSpvInyHfPG:-OoQV_WmCo7QkELxVZf9:b-v5t8mb) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/ORelayServer.cpp#L69)):*

We need the executor for aggregation. Now that the collector_ is bound by the ORelayServer, the keepAlive reference would get destroyed with the server. 
Do you think we need an explicit shutdown tied to MoQServer::stop?
</blockquote></details>

Maybe?  The problem will present itself later, if there is one.
___
*[`src/admin/MetricsHandler.cpp` line 28 at r1](https://reviewable.io/reviews/openmoq/o-rly/49#-OnTYzEv-DzgnzDzaQ8D:-OoQWGdwCVQKxeRH8t3o:begtiu8) ([raw file](https://github.com/openmoq/o-rly/blob/22495123caa8921660d6b2117a135d80ad64022e/src/admin/MetricsHandler.cpp#L28)):*

modified to use `folly::EventBaseManager::get()->getEventBase();` for simplicity
</blockquote></details>

I think this is right -- the lambda that generates the response _has_ to run in the transaction's EVB.
___
*[`include/o_rly/admin/MetricsHandler.h` line 19 at r3](https://reviewable.io/reviews/openmoq/o-rly/49#-OoQSdGh5z1nAOepkuf4:-OoQSdGh5z1nAOepkuf5:b-ivinl1) ([raw file](https://github.com/openmoq/o-rly/blob/3b00270655beca486fd5d9c86aa4667564021289/include/o_rly/admin/MetricsHandler.h#L19)):*
> ```objc
> // Registers GET /metrics on the given admin server.
> //
> // The handler aggregates stats asynchronously using co_withCancellation so
> ```

This comment may have too much detail for where it sits.
___
*[`include/o_rly/stats/MoQStatsCollector.h` line 28 at r3](https://reviewable.io/reviews/openmoq/o-rly/49#-OoQZmAq3U-TNervEydx:-OoQZmAq3U-TNervEydy:b-gemquh) ([raw file](https://github.com/openmoq/o-rly/blob/3b00270655beca486fd5d9c86aa4667564021289/include/o_rly/stats/MoQStatsCollector.h#L28)):*
> ```objc
> public:
>   // --- Inner callback: publisher role ---
>   class PublisherCallback : public moxygen::MoQPublisherStatsCallback {
> ```

This may be more a moxygen question -- how are the Publisher and Subscriber stats callbacks different?

I wonder if the underlying counters are the same (eg subscribeSuccess) can we share more code?  Base class with counters common to sub and pub?
___
*[`include/o_rly/stats/StatsRegistry.h` line 26 at r3](https://reviewable.io/reviews/openmoq/o-rly/49#-OoQUBznCsqZcivMkQDq:-OoQUBznCsqZcivMkQDr:bfqhvvo) ([raw file](https://github.com/openmoq/o-rly/blob/3b00270655beca486fd5d9c86aa4667564021289/include/o_rly/stats/StatsRegistry.h#L26)):*
> ```objc
> // small ordered lookup table and a constexpr mapping function.  Unknown codes
> // fall into the last slot (CANCELLED).
> inline constexpr std::array<moxygen::RequestErrorCode, 8> kRequestErrorCodeValues = {{
> ```

It's a little annoying to keep this in sync.  Can you add a static_assert that the size of the array matches the number of enums in MoQTypes?
___
*[`include/o_rly/stats/StatsRegistry.h` line 41 at r3](https://reviewable.io/reviews/openmoq/o-rly/49#-OoQUO8g2fMX46kyJkWq:-OoQUO8g2fMX46kyJkWr:bwz6prs) ([raw file](https://github.com/openmoq/o-rly/blob/3b00270655beca486fd5d9c86aa4667564021289/include/o_rly/stats/StatsRegistry.h#L41)):*
> ```objc
> // Maps any RequestErrorCode to a compact index in [0, kRequestErrorCodeCount).
> // Unknown / future codes fall into the last slot (CANCELLED).
> constexpr size_t requestErrorCodeIndex(moxygen::RequestErrorCode code) {
> ```

I don't love this linear search (which is on every stat bump, but just errors I guess). If the range is compact we can just cast to integer?

If you want to roll with what you have, just file an issue?
___
*[`src/ORelayServer.cpp` line 65 at r3](https://reviewable.io/reviews/openmoq/o-rly/49#-OoQV0bKCOruqXCnpw1N:-OoQV0bKCOruqXCnpw1O:b-wi8jzg) ([raw file](https://github.com/openmoq/o-rly/blob/3b00270655beca486fd5d9c86aa4667564021289/src/ORelayServer.cpp#L65)):*
> ```cpp
>   if (statsRegistry_) {
>     if (!statsCollector_) {
>       statsCollector_ = stats::MoQStatsCollector::create_moq_stats_collector(
> ```

Maybe just add a note that we need a statsCollector per exec, and we'll expand this when we go multi-threaded.
___
*[`src/admin/MetricsHandler.cpp` line 36 at r3](https://reviewable.io/reviews/openmoq/o-rly/49#-OoQWu-61FFwQb2KxRJ1:-OoQWu-61FFwQb2KxRJ2:bi2xj18) ([raw file](https://github.com/openmoq/o-rly/blob/3b00270655beca486fd5d9c86aa4667564021289/src/admin/MetricsHandler.cpp#L36)):*
> ```cpp
>                   stats::StatsSnapshot snap;
>                   try {
>                     snap = co_await reg->aggregateAsync();
> ```

Oh and aggregateAsync handles dispatch to executor threads it seems
___
*[`src/admin/MetricsHandler.cpp` line 36 at r3](https://reviewable.io/reviews/openmoq/o-rly/49#-OoQXFWSCFvgo5q_4tbI:-OoQXFWSCFvgo5q_4tbJ:bt2fgoy) ([raw file](https://github.com/openmoq/o-rly/blob/3b00270655beca486fd5d9c86aa4667564021289/src/admin/MetricsHandler.cpp#L36)):*
> ```cpp
>                   stats::StatsSnapshot snap;
>                   try {
>                     snap = co_await reg->aggregateAsync();
> ```

An alternate pattern is

```
auto snap = co_await co_awaitTry(reg->aggregateAsync());
if (snap.hasException()) {
  ...
}
```

then, you don't nee try/catch.  But it's more a style issue - your choice.
___
*[`src/stats/MoQStatsCollector.cpp` line 16 at r3](https://reviewable.io/reviews/openmoq/o-rly/49#-OoQZ3Ir-snAQHfcgWpX:-OoQZ3Ir-snAQHfcgWpY:byzmqtg) ([raw file](https://github.com/openmoq/o-rly/blob/3b00270655beca486fd5d9c86aa4667564021289/src/stats/MoQStatsCollector.cpp#L16)):*
> ```cpp
>   collector->registry_ = registry;
> 
>   // Build aliased shared_ptrs: they share the parent's refcount but point to
> ```

I have never used this, but it looks clever

afrind (approved):
> @afrind resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n).

akash-a-n (commented):
> @akash-a-n made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/o-rly/49)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n).

___
*[`src/ORelayServer.cpp` line 69 at r2](https://reviewable.io/reviews/openmoq/o-rly/49#-OnZ8KNI3FSpvInyHfPG:-OoZxrQo2Se2rFoUrHL9:b-l5ywn5) ([raw file](https://github.com/openmoq/o-rly/blob/5c13d4cd04a5248b1fd03cb2f70591a1cbce274f/src/ORelayServer.cpp#L69)):*

Maybe?  The problem will present itself later, if there is one.
</blockquote></details>

Will take care of this in the upcoming PRs

PR #80, opened 2026-03-27
Upstream provider for relay-to-relay

afrind:
Still a WIP, but tested with 2 relays for publishNamespace+subscribe flows.

1. Each Relay can declare a string relay ID, if unset, randomly generated
2. Each service can optionally list an upstream URL
3. moqx will proactively establish a draft-16 session the the upstream
4. It will auto-reconnect with exponential backoff if it breaks/fails to connect
5. It sends a SUBSCRIBE_NAMESPACE(*, BOTH) to upstream including it's relay ID in an Auth Token
6. When receiving a SUBSCRIBE_NAMESPACE with the Auth Token, send a reciprocal SUBSCRIBE_NAMESPACE(*, BOTH) [no token]
7. Any PUBLISH entering the network will automatically flow through all chained relays to end subscribers
8. Any PUBLISH_NAMESPACE entering the network will flow through chained relays as NAMESPACE
9. Any SUBSCRIBE, FETCH or TRACK_STATUS will automatically follow the chain to the publisher
10. SUBSCRIBE_NAMESPACE does not need to be propagated, since there's already a maximal one to every peer

Limitations:

1. If you configure a loop, e.g. `A <-> B` or `A -> B -> C -> A`, you're gonna have a bad time.  Stick to tree-structures.
2. Multipublisher is not yet supported, and last publisher wins.  Races are possible and not pretty.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/80)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,6 +258,15 @@ if(MOQX_BUILD_TESTS)
   gtest_discover_tests(moqx_service_matcher_test)
 
   add_test(
+    NAME relay_chain
+    COMMAND bash ${PROJECT_SOURCE_DIR}/tests/test_relay_chain.sh $<TARGET_FILE:moqx>
+  )
+  set_tests_properties(relay_chain PROPERTIES
+    TIMEOUT 120
+    ENVIRONMENT "MOQBIN=${PROJECT_SOURCE_DIR}/.scratch/moxygen-install/bin"
+  )
+
+  add_test(
     NAME admin_info_endpoint
     COMMAND bash ${PROJECT_SOURCE_DIR}/tests/test_admin_info.sh $<TARGET_FILE:moqx>
   )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,7 @@ add_library(moqx_core STATIC
   src/stats/StatsRegistry.cpp
   src/stats/MoQStatsCollector.cpp
   src/stats/QuicStatsCollector.cpp
+  src/UpstreamProvider.cpp
 )
 
 target_include_directories(moqx_core
@@ -98,6 +99,7 @@ target_link_libraries(moqx_core PUBLIC
   # Workaround: wangle::wangle_acceptor_acceptor_core uses AsyncFdSocket from
   # FizzAcceptorHandshakeHelper but omits this dep from its cmake config.
   Folly::folly_io_async_fdsock_async_fd_socket
+  moxygen::moxygen_moqclient
 )
 
 target_compile_options(moqx_core PRIVATE -Wall -Wextra -Wpedantic)
@@ -267,6 +269,20 @@ if(MOQX_BUILD_TESTS)
     NAME admin_metrics_endpoint
     COMMAND bash ${PROJECT_SOURCE_DIR}/tests/test_admin_metrics.sh $<TARGET_FILE:moqx>
   )
+
+  add_executable(moqx_upstream_provider_test
+    tests/UpstreamProviderTest.cpp
+  )
+  target_link_libraries(moqx_upstream_provider_test PRIVATE
+    moqx_core
+    moqx_test_utils
+    moqx_test_main
+    moxygen::moqtest_utils
+    moxygen::moxygen_events_moq_folly_executor_impl
+    moxygen::moxygen_moqclient
+    GTest::gmock
+  )
+  gtest_discover_tests(moqx_upstream_provider_test)
 endif()
 
 include(${PROJECT_SOURCE_DIR}/cmake/Lint.cmake)
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -35,6 +35,11 @@ services:
       max_tracks: 500             # Override: 100 → 500
       max_groups_per_track: 10    # Override: 3 → 10
       # enabled: inherited (true) from service_defaults
+    # upstream:                   # Optional: enable relay chaining for this service
+    #   url: "moqt://origin.example.com:4433/moq-relay"
+    #   tls:
+    #     insecure: false         # true = skip cert verification (dev only)
+    #     # ca_cert: /path/to/ca.pem  # Custom CA cert (mutually exclusive with insecure: true)
 
   testing:
     match:
@@ -53,3 +58,5 @@ admin:
   # tls:
   #   cert_file: /path/to/cert.pem
   #   key_file: /path/to/key.pem
+
+# relay_id: "my-relay-1"     # Relay identity (optional; random hex string generated if absent)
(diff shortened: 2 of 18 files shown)


wilaw:
This design constrains the network to a linear topology, as shown in A. Blue and orange represent different subnets (CDNs).  If a client publishes into node 1 and the end-subscriber is attached to node 7, then the traffic must traverse through relays 2,3,4,5 and 6. This is quite inefficient, even for a simple, get-it-working quickly approach. 

What if you extended the scheme to allow the listing N upstream URLs? To prevent loops and gain a simple mechanism for path selection , we can implement a path vectoring scheme, as is done with BGP. 

When a node originates or forwards a discovery/subscription advertisement, it appends its relay ID to a list within the message. We can use the token to transmit this list, or else define a custom parameter which accompanies the SUBSCRIBE_NAMESPACE. 

Originating Node: When Node 1 generates an advertisement, it sets the path to [1].

Forwarding Node: When Node 3 receives an advertisement from Node 1, it checks if its own ID (3) is already in the path list. If not, it appends its ID and forwards the message to Node 5 with the path [1, 3].

Loop Detection: If Node 5 forwards that message to Node 7, and Node 7 attempts to send it back to a neighbor that eventually reaches Node 3, Node 3 will see its own ID (3) already present in the path list [1, 3, 5, 7].

Action: If a node sees its own ID in the path of an incoming discovery message, **it must silently discard the message**.

Upstream Selection: In Topology B, Node 5 now has multiple potential upstreams (3, 6, 7, 8). By looking at the path_vector, Node 5 can also calculate the "shortest" path (the one with the fewest hops) to the origin, allowing it to choose the most efficient upstream rather than just the first one it hears from.

Advantages:
1. No Global State: It doesn't require a central "topology manager" or a complex Link-State DB.
2. Local Logic: Each relay only needs to know its own ID and look at the incoming message to make a "drop or forward" decision.
3. Subnet-Aware: This handles the transition between the blue and orange subnets , as the path vector will show the crossing of the boundary.
4. Shortest path selection when there is a choice in upstreams. 

Screenshot 2026-03-27 at 10 03 58

afrind (commented):
Discussed in #relay-devs -- the current design is not final, and supports non-linear topologies (eg trees) -- just not full-mesh/loops

> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/80#-:-OpBZPBu7vaB5VkEqyfC:b-n9eij0)* status: 0 of 31 files reviewed, all discussions resolved.

afrind (commented):
I see the ORelay/o-rly references and have fixed locally

> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/80#-:-OpBaRRk3Lh9dXgMK1CD:b-9hnrjz)* status: 0 of 31 files reviewed, all discussions resolved.

akash-a-n (commented):
> @akash-a-n reviewed 2 files and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/80)* status: 2 of 31 files reviewed, 3 unresolved discussions (waiting on afrind).

___
*[`.gitmodules` line 4 at r3](https://reviewable.io/reviews/openmoq/moqx/80#-OpCdT3q1ZQPs_fcINRT:-OpCdT3q1ZQPs_fcINRU:brjijer) ([raw file](https://github.com/openmoq/moqx/blob/c523385f5d189d07fd28c8fab868be7ba84b58dc/.gitmodules#L4)):*
> ```gitconfig
> 	path = deps/moxygen
> 	url = git@github.com:openmoq/moxygen.git
> 	branch = subscribe-ns-fix
> ```

is this intended?
___
*[`src/UpstreamProvider.cpp` line 51 at r3](https://reviewable.io/reviews/openmoq/moqx/80#-OpDFxjn77xqPlYgxFZF:-OpDFxjn77xqPlYgxFZG:bvh4ae8) ([raw file](https://github.com/openmoq/moqx/blob/c523385f5d189d07fd28c8fab868be7ba84b58dc/src/UpstreamProvider.cpp#L51)):*
> ```cpp
> 
> // TrackConsumer proxy whose downstream is set after connecting to upstream.
> // Per MoQ protocol, publishers must not send data before PublishOk, so
> ```

per protocol, publishers can send data before publish_ok when forward=1/absent. 
So we might need a trackConsumer implementation that asks the original publisher to wait or sets forward=0?
___
*[`src/UpstreamProvider.cpp` line 176 at r3](https://reviewable.io/reviews/openmoq/moqx/80#-OpDFBS29IIdt-R0peo7:-OpDFBS29IIdt-R0peo8:b-xulrod) ([raw file](https://github.com/openmoq/moqx/blob/c523385f5d189d07fd28c8fab868be7ba84b58dc/src/UpstreamProvider.cpp#L176)):*
> ```cpp
> }
> 
> folly::coro::Task<Publisher::SubscribeResult>
> ```

nice optimization.

afrind (commented):
> @afrind made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/80)* status: 2 of 31 files reviewed, 2 unresolved discussions (waiting on akash-a-n).

___
*[`.gitmodules` line 4 at r3](https://reviewable.io/reviews/openmoq/moqx/80#-OpCdT3q1ZQPs_fcINRT:-OpDLuZi4f0W70s_LRof:b-dv0s0m) ([raw file](https://github.com/openmoq/moqx/blob/c523385f5d189d07fd28c8fab868be7ba84b58dc/.gitmodules#L4)):*

is this intended?
</blockquote></details>

yes for now, there's a moxygen bug in subscribeNamespace.  See https://github.com/facebookexperimental/moxygen/pull/126 
___
*[`src/UpstreamProvider.cpp` line 51 at r3](https://reviewable.io/reviews/openmoq/moqx/80#-OpDFxjn77xqPlYgxFZF:-OpDM9Cb7sKMLu7qxSz8:bngcwx9) ([raw file](https://github.com/openmoq/moqx/blob/c523385f5d189d07fd28c8fab868be7ba84b58dc/src/UpstreamProvider.cpp#L51)):*

per protocol, publishers can send data before publish_ok when forward=1/absent. 
So we might need a trackConsumer implementation that asks the original publisher to wait or sets forward=0?
</blockquote></details>

Dang, we need to give some more thought to this publish flow.  The PendingTrackConsumer may actually need to buffer objects while waiting for the connect to go through.

akash-a-n (commented):
> @akash-a-n partially reviewed 9 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/80)* status: 11 of 31 files reviewed, all discussions resolved.

akash-a-n (approved):
:lgtm:

> @akash-a-n made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/80#-:-OpGiDGd8lWnOj1-DF8p:bnfp4nl)* status: 11 of 31 files reviewed, all discussions resolved.

PR #98, opened 2026-03-31
Add --profile san support to from-source dep build

afrind:
Passes --profile to setup-deps-standalone.sh, which uses a separate build/install dir (standalone-build-san, moxygen-install-san) and compiles deps with ASAN. UBSAN is excluded from deps because folly's static_assert on ::recvmmsg/::sendmmsg becomes non-constant under UBSAN's function interposition. The moqx san preset retains full ASAN+UBSAN as before.

cmd_build prefers cmake_prefix_path-<profile>.txt when present, falling back to the default prefix path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/98)

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -6,7 +6,7 @@
 #   --from-source   — build moxygen + all Meta deps from source (~15-30 min)
 #
 # Usage:
-#   ./scripts/build.sh setup [--from-release [SHA]|--from-source [SHA]] [--no-fallback] [--clean]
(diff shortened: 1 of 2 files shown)


gmarzot (commented):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/98)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #115, opened 2026-04-05
Add micro-benchmark suite (folly Benchmark)

gmarzot:
## Summary

Add micro-benchmark suite using folly Benchmark, modeled after Quicr/libquicr.

- **9 benchmark files, 63 benchmarks** covering:
  - **MOQT framer**: varint encode/decode (cold + warm), frame writes and parses (Subscribe, Subgroup, StreamObject, PublishNamespace, Fetch, Goaway), Subscribe roundtrip, TrackNamespace/FullTrackName operations
  - **MOQT extensions**: serialize, deserialize, and roundtrip at 1/10/100/1000 extensions (direct comparison to libquicr)
  - **Stats**: collector callbacks, snapshot, histogram, Prometheus text formatting at scale
  - **Service matcher**: exact/wildcard/fallback/no-match routing
  - **Config**: YAML load, resolve, schema generation
- folly Benchmark via the existing standalone folly fetch — **no new top-level dep**. Links `Folly::follybenchmark` (already built as part of the moxygen standalone build).
- `--benchmark` flag for `scripts/build.sh`
- CI benchmark jobs on ubuntu-22.04 and macos-latest
- macOS gflags shared linking fix (`GFLAGS_SHARED=ON` on Darwin)
- macOS cmake version check fix (`grep -P` → `sed`)

## Build & run

```
./scripts/build.sh --benchmark
./build/benchmark/moqx_benchmark
```

## Test plan
- [x] Full suite passes locally on WSL2 (linux x86)
- [x] Full suite passes on argo (macOS ARM64)
- [x] CI benchmark passes on ubuntu-22.04
- [x] CI benchmark passes on macos-latest
- [ ] Same-hardware comparison with libquicr on argo (see benchmark report comment)

## Future benchmarks
- Relay forwarding throughput (equivalent to libquicr's `PQ_ConnDataForwarding`, requires mock sessions)
- Subgroup object header parse (hot path for object delivery)
- Multi-session stats aggregation (`StatsRegistry::aggregateAsync`)

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/115)

PR #124, opened 2026-04-07
Add macOS build to CI

gmarzot:
## Summary
Add `macos-latest` to the ci-pr build matrix so moqx is built and tested on macOS.

## Changes
- Add `macos-latest` entry to ci-pr.yml build matrix
- Fix `install-system-deps.sh` invocation (no `sudo` on macOS)
- Fix `grep -P` → `sed` for cmake version check (macOS `grep` lacks `-P`)
- Add `GFLAGS_SHARED=ON` on Darwin to avoid static/shared gflags conflict between brew and moxygen tarball (see openmoq/moxygen#114)

## Dependencies
- Overlaps with PR #115 (benchmark) which has the same `build.sh` fixes. Whichever merges first, the other will need a trivial rebase.

## Known issues
- macOS test discovery may fail for `moqx_config_test` due to gflags conflict at test enumeration time (same issue #115 solved by using `MOQX_BUILD_TESTS=OFF` for benchmarks). If tests fail, we may need `GFLAGS_SHARED=ON` set at the CMakeLists level rather than just in build.sh.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/124)

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -63,6 +63,8 @@ jobs:
         run: |
           if [[ "$(uname)" == "Darwin" ]]; then
             deps/moxygen/standalone/install-system-deps.sh
+            brew install coreutils
+            echo "/opt/homebrew/opt/coreutils/libexec/gnubin" >> "$GITHUB_PATH"
           else
             sudo deps/moxygen/standalone/install-system-deps.sh
           fi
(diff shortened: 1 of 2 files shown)


afrind (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/124)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gmarzot).

PR #135, opened 2026-04-09
pico: enable WebTransport and wire WT paths from service config

afrind:
Pico listeners now run with WebTransport over H3 enabled (enableWebTransport=true, enableQuicTransport=true, wtMaxSessions=1). The h3zero WT path table is populated from the exact-path service match entries via ServiceMatcher::allExactPaths(), so no additional YAML config is needed — the service definitions already encode which paths clients will CONNECT to.

Config validator warns when a pico listener has services with prefix-path match rules, since h3zero routes WebTransport CONNECT by prefix-up-to-'?' matching only; sub-path routing is not supported and clients matching only a prefix rule may fail to connect.

Also bumps deps/moxygen to feature/pico-wt-multi-path which extends PicoWebTransportConfig.wtEndpoints to a vector and sizes the h3zero path table dynamically.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/135)

### DIFF
--- a/src/MoqxPicoRelayServer.cpp
+++ b/src/MoqxPicoRelayServer.cpp
@@ -69,7 +69,14 @@ MoqxPicoRelayServer::MoqxPicoRelayServer(
           listenerCfg.endpoint,
           ioExecutor->getAllEventBases()[0],
           listenerCfg.moqtVersions,
-          picoTransportConfigFromQuicConfig(listenerCfg.quic)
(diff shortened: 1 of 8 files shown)


suhasHere (approved):
> @suhasHere reviewed 9 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/135)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #144, opened 2026-04-09
fix: null deref in publish() when publisher reconnects with draining subscriber

afrind:
When a publisher terminates while a subscriber has an open subgroup, onPublishDone() resets handle to null but keeps the subscriptions_ entry alive (the forwarder marks the subscriber receivedPublishDone_ and waits for subgroups to close). If the publisher then reconnects and re-publishes the same track, the multipublisher path calls handle->unsubscribe() on the null handle — SIGSEGV.

Guard both handle->unsubscribe() and forwarder->publishDone() on the null check: when handle is null the previous publisher already terminated and the forwarder was already drained, so both calls are wrong.

Adds a regression test that reproduces the crash by opening a subgroup before dropping the publisher connection.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/144)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -460,13 +460,20 @@ MoqxRelay::publish(PublishRequest pub, std::shared_ptr<Publisher::SubscriptionHa
     // multipublisher
     XLOG(DBG1) << "New publisher for existing subscription";
     nodePtr->publishes.erase(pub.fullTrackName.trackName);
-    it->second.handle->unsubscribe();
-    it->second.forwarder->publishDone(
-        {RequestID(0),
-         PublishDoneStatusCode::SUBSCRIPTION_ENDED,
-         0, // filled in by session
-         "upstream disconnect"}
-    );
+    // handle is null when the previous publisher already terminated (e.g.
+    // reconnect after a dropped connection with subscribers still draining open
+    // subgroups).  onPublishDone() already reset handle and drained the
+    // forwarder, so skip both calls to avoid a null deref and a double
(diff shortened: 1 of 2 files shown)


src/MoqxRelay.cpp
```diff
@@ -86,7 +86,8 @@ MoqxRelay::doSubscribeUpdate(std::shared_ptr<Publisher::SubscriptionHandle> hand
        /*forward=*/forward}
   );
   if (updateRes.hasError()) {
-    XLOG(ERR) << "requestUpdate failed: " << updateRes.error().reasonPhrase;
+    XLOG(ERR) << "requestUpdate failed: " << updateRes.error().reasonPhrase << ", unsubscribing";
+    handle->unsubscribe();
```
afrind: Dang it this is part of another fix I was doing in parallel

peterchave (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/144#-:-Opo8vrwDS_xxZ4gQhRv:bnfp4nl)* status: all files reviewed, 1 unresolved discussion.

src/MoqxRelay.cpp
peterchave: LGTM

peterchave (approved):
> @peterchave made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/144)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #146, opened 2026-04-09
admin: add GET /state endpoint with relay state visitor pattern

afrind:
Adds a /state JSON endpoint to the admin server that dumps live relay state: active sessions, downstream peers (with relay ID), per-track subscriptions (subscriber counts, counters, largest), namespace announcement tree, and cache stats.

Design uses a visitor pattern (RelayStateVisitor / RelayContextVisitor) so MoqxRelay and MoqxRelayContext have no JSON dependency — all serialization lives in StateHandler.cpp via a streaming JsonWriter backed by folly::IOBufQueue.

Also:
- getPeerRelayID() replaces isPeerSubNs(), extracting the peer relay ID from the auth token so it appears in /state output
- UpstreamProvider::stateString() exposes upstream connection state
- test_relay_chain.sh gains check_state() and /state assertions after Direction 1 (peer count, namespace tree, upstream.state=connected)

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/146)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,7 @@ add_library(moqx_core STATIC
   src/admin/AdminServer.cpp
   src/admin/BuiltinRoutes.cpp
   src/admin/MetricsHandler.cpp
+  src/admin/StateHandler.cpp
   src/stats/StatsRegistry.cpp
   src/stats/MoQStatsCollector.cpp
   src/stats/PicoQuicStatsCollector.cpp
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -735,8 +735,9 @@ folly::coro::Task<Publisher::SubscribeNamespaceResult> MoqxRelay::subscribeNames
   // Relay peering: if the incoming subNs carries a relay auth token, the peer
   // is a relay. Reciprocate with our own peer subNs so the peer gets our
   // namespace announcements as publishers connect.
-  if (!relayID_.empty() && isPeerSubNs(subNs)) {
-    XLOG(INFO) << __func__ << ": peer relay detected, reciprocating peer subNs";
+  if (auto peerID = !relayID_.empty() ? getPeerRelayID(subNs) : std::nullopt) {
+    XLOG(INFO) << __func__ << ": peer relay detected peer_id=" << *peerID
+               << ", reciprocating peer subNs";
     auto handle = makeNamespaceBridgeHandle(weak_from_this(), session);
     auto recipResult = co_await session->subscribeNamespace(
         makePeerSubNs(),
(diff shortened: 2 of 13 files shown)


peterchave:
*[`src/MoqxRelay.cpp` line 1175 at r1](https://reviewable.io/reviews/openmoq/moqx/146#-Oprj2WeASXuy1BcQ8DF:-Oprj2WeASXuy1BcQ8DG:blk4hj3) ([raw file](https://github.com/openmoq/moqx/blob/0efb8f9cc29ae3ca846d430186f9cd30b6ba04da/src/MoqxRelay.cpp#L1175)):*
> ```cpp
>   visitor.onPeersBegin();
>   for (const auto& [sess, peer] : peerSubNsHandles_) {
>     visitor.onPeer(sess->getPeerAddress().describe(), sess->getAuthority(), peer.relayID);
> ```

For Relay IDs - is this the user configured name set in the config or internal id?
If the user configured name - what would happen if two relays were set to same name?
Would that just make the report confusing - or does that cause issues?

peterchave:
*[`src/admin/StateHandler.cpp` line 235 at r1](https://reviewable.io/reviews/openmoq/moqx/146#-OprmNhr0v6VoCpa5-pL:-OprmNhr0v6VoCpa5-pM:b-r8mjg8) ([raw file](https://github.com/openmoq/moqx/blob/0efb8f9cc29ae3ca846d430186f9cd30b6ba04da/src/admin/StateHandler.cpp#L235)):*
> ```cpp
>                   }
>                   std::unique_ptr<folly::IOBuf> body;
>                   try {
> ```

For future consideration, do we need a timeout option? 
For a heavily loaded server, this could generate a lot of data... would the times takes to do that be a concern?

For a future implementation also might want to rate limit this API anyway?

peterchave (approved):
:lgtm:

> @peterchave reviewed 14 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/146#-:-OpsAhTg4cwHi-gU6AGc:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on afrind).

afrind (commented):
> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/146)* status: 5 of 14 files reviewed, 1 unresolved discussion (waiting on akash-a-n, gmarzot, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

___
*[`src/MoqxRelay.cpp` line 1175 at r1](https://reviewable.io/reviews/openmoq/moqx/146#-Oprj2WeASXuy1BcQ8DF:-Oq8qDVn1aJZTLmS5vDP:b91qox) ([raw file](https://github.com/openmoq/moqx/blob/0efb8f9cc29ae3ca846d430186f9cd30b6ba04da/src/MoqxRelay.cpp#L1175)):*
<details><summary><i>Previously, peterchave wrote…</i></summary><blockquote>

For Relay IDs - is this the user configured name set in the config or internal id?
If the user configured name - what would happen if two relays were set to same name?
Would that just make the report confusing - or does that cause issues?
</blockquote></details>

This is a user configured name.  It's not critical for lookups at this point, just informational (for this /state dump, or logs).  Once we have more of a real mesh, we'd use it for loop detection and match selection, etc.

PR #148, opened 2026-04-10
fix: restore MoQServer optimized defaults and enable pacing in buildTransportSettings

afrind:
Supplying a TransportSettings to MoQServer bypasses its optimized defaults (GSO batching, ContinuousMemory data path, large cwnd, etc.). Replicate those defaults explicitly before applying config overrides so nothing is silently lost.

Also enables pacingEnabled unconditionally — BBR requires it (mvfst falls back to Cubic without it), and pacing is beneficial for all supported CC algorithms.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/148)

### DIFF
--- a/src/MoqxRelayServer.cpp
+++ b/src/MoqxRelayServer.cpp
@@ -54,7 +54,17 @@ buildFizzContext(const config::ListenerConfig& cfg) {
 }
 
 quic::TransportSettings buildTransportSettings(const config::QuicConfig& quic) {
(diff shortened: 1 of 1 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/148)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #158, opened 2026-04-11
MoqxRelay: fix TOCTOU race in subscribe() causing PromiseAlreadySatisfied crash

afrind:
Two concurrent coroutines both passed subscriptions_.find()==end() before either emplaced, because the only suspension between the find() and emplace() was co_await upstream_->waitForConnected().  Both then called rsub.promise.setValue() on the same entry, throwing PromiseAlreadySatisfied → std::terminate.

Fix: after waitForConnected() returns, re-check subscriptions_.find().  If another coroutine emplaced the entry while we were suspended, update subscriptionIt and fall through to the second-subscriber path instead of proceeding as first subscriber.

Adds regression test RelayUpstreamSubscribeRaceTest::ConcurrentSubscribesSameTrack that reproduces the crash without the fix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/158)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -852,6 +852,18 @@ folly::coro::Task<Publisher::SubscribeResult>
 MoqxRelay::subscribe(SubscribeRequest subReq, std::shared_ptr<TrackConsumer> consumer) {
   auto session = MoQSession::getRequestSession();
   auto subscriptionIt = subscriptions_.find(subReq.fullTrackName);
+  // TOCTOU fix: if we might be the first subscriber, wait for the upstream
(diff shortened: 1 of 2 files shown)


suhasHere (approved):
> @suhasHere reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/158)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/158)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #159, opened 2026-04-11
Apply upstream relay bug fixes

afrind:
Three use-after-free / null-deref fixes ported from moxygen MoQRelay.cpp:
- publish(): move forwarder out and erase subscription entry before calling publishDone, preventing UAF when drainSubscriber fires onEmpty mid-iteration
- fetch(): check !upstreamSession after subscription lookup so a null upstream (publisher terminated, entry still live) returns TRACK_NOT_EXIST instead of crashing
- forwardChanged(): guard against null handle after publisher termination so a requestUpdate(forward=false) arriving after onPublishDone doesn't deref null upstream

Five regression tests added to MoqxRelayTest.cpp covering all three fixes.

Bumps moxygen submodule to openmoq/moxygen sync/pr-141-fixes (5232364b), which contains the MoQForwarder null-trackConsumer fix the new tests depend on.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/159)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -481,22 +481,29 @@ MoqxRelay::publish(PublishRequest pub, std::shared_ptr<Publisher::SubscriptionHa
     // multipublisher
     XLOG(DBG1) << "New publisher for existing subscription";
     nodePtr->publishes.erase(pub.fullTrackName.trackName);
+    // Move the forwarder out and erase the entry BEFORE calling publishDone.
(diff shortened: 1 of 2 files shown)


suhasHere (approved):
> @suhasHere reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/159)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #167, opened 2026-04-11
feat: Wire MoqxRelay self-exclusion for publisher-subscribers

afrind:
When a session publishes a track while also holding a TRACK_FILTER subscription, that track is registered as a self-track in PropertyRanking so the publisher never receives its own stream in their top-N selection.

Three integration points:

publish(): addPublishedTrackToSession is called before registerTrack so
  that notifications fired during registration already see the self-track
  exclusion, preventing the brief window where a publisher would receive
  its own track.

subscribeNamespace(): tracks already published by this session are passed
  to addSessionToTopNGroup as the initial self-exclusion set, covering the
  case where a session subscribes after it has already been publishing.

onPublishDone(): removePublishedTrackFromSession is called before erasing
  from the publishes map so the publisher's selection can be reconciled
  and any replacement track notified.

PropertyRanking::registerTrack also gains an else-branch that reconciles publisher-subscriber sessions for tracks landing outside the shared top-N: a publisher's personal top-N extends further than the shared top-N (self- tracks consume slots), so tracks at rank >= N may still be in their view.

Adds three publisher-subscriber integration tests to MoqxTrackFilterTest: subscribe-before-publish, publish-before-subscribe, and viewer-sees-self- track-while-publisher-does-not.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/167)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,7 @@ add_library(moqx_core STATIC
   src/stats/QuicStatsCollector.cpp
   src/UpstreamProvider.cpp
   src/relay/TopNFilter.cpp
+  src/relay/PropertyRanking.cpp
 )
 
 target_include_directories(moqx_core
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -412,21 +412,9 @@ void MoqxRelay::onPublishDone(const FullTrackName& ftn) {
       // Remove from publishes map
       auto nodePtr = findNamespaceNode(ftn.trackNamespace);
       if (nodePtr) {
-        // If the publisher also had a TRACK_FILTER subscription, remove the
-        // self-track from their exclusion set before erasing from publishes.
-        auto publishIt = nodePtr->publishes.find(ftn.trackName);
-        if (publishIt != nodePtr->publishes.end()) {
-          auto& publisherSession = publishIt->second;
-          auto sessIt = nodePtr->sessions.find(publisherSession);
-          if (sessIt != nodePtr->sessions.end() && sessIt->second.trackFilter) {
-            auto& tf = *sessIt->second.trackFilter;
-            auto rankingIt = nodePtr->rankings.find(tf.propertyType);
-            if (rankingIt != nodePtr->rankings.end()) {
-              rankingIt->second
-                  ->removePublishedTrackFromSession(tf.maxSelected, publisherSession, ftn);
-            }
-          }
-        }
+        // Note: Self-exclusion cleanup happens automatically when removeTrack()
+        // is called (via TopNFilter's onTrackEnded observer). The publisher's
+        // personal selection is reconciled inside PropertyRanking.
         bool hadLocalContent = nodePtr->hasLocalSessions();
         nodePtr->publishes.erase(ftn.trackName);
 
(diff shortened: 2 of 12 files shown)


suhasHere (approved):
> @suhasHere reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/167)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

afrind (commented):
> @afrind made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/167)* status: all files reviewed, 4 unresolved discussions (waiting on suhasHere).

___
*[`src/MoqxRelay.cpp` line 426 at r1](https://reviewable.io/reviews/openmoq/moqx/167#-Oq1dN3X2lNcoHIksHy5:-Oq1dN3X2lNcoHIksHy6:bpprnui) ([raw file](https://github.com/openmoq/moqx/blob/32b594bb45731dbed0fb03f8cfd709d1b654882c/src/MoqxRelay.cpp#L426)):*
> ```cpp
>             if (rankingIt != nodePtr->rankings.end()) {
>               rankingIt->second
>                   ->removePublishedTrackFromSession(tf.maxSelected, publisherSession, ftn);
> ```

This is where the API 169 is called with a specific N but it's still odd.
___
*[`src/relay/PropertyRanking.cpp` line 56 at r1](https://reviewable.io/reviews/openmoq/moqx/167#-Oq1dYd63mVyq75pr3AQ:-Oq1dYd63mVyq75pr3AR:bkxi9k2) ([raw file](https://github.com/openmoq/moqx/blob/32b594bb45731dbed0fb03f8cfd709d1b654882c/src/relay/PropertyRanking.cpp#L56)):*
> ```cpp
>       // If that track was Selected, demote it into the deselected queue.
>       demoteTrackAtRank(n, topNGroup);
>     } else {
> ```

Ooops this should go in 169 along with a test.  167 should be relay only.
___
*[`test/MoqxTrackFilterTest.cpp` line 788 at r1](https://reviewable.io/reviews/openmoq/moqx/167#-Oq1dhR4AixjvBhA9kqg:-Oq1dhR56TWDCqm2BIPF:b-rmgz80) ([raw file](https://github.com/openmoq/moqx/blob/32b594bb45731dbed0fb03f8cfd709d1b654882c/test/MoqxTrackFilterTest.cpp#L788)):*
> ```cpp
>   auto pub = makeSession();
>   auto other = makeSession();
>   doPublishNamespace(pub);
> ```

cargo-cult
___
*[`test/MoqxTrackFilterTest.cpp` line 791 at r1](https://reviewable.io/reviews/openmoq/moqx/167#-Oq1doU_DY79-mEyUZWq:-Oq1doU_DY79-mEyUZWr:b-m2ir9r) ([raw file](https://github.com/openmoq/moqx/blob/32b594bb45731dbed0fb03f8cfd709d1b654882c/test/MoqxTrackFilterTest.cpp#L791)):*
> ```cpp
> 
>   // Subscribe before publishing.
>   doSubscribeFilter(pub, /*maxSelected=*/2);
> ```

I think the test is more compelling if you drive here and verify you get a and b, then publish self and observe no change.

suhasHere (approved):
> @suhasHere reviewed 13 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/167)* status: all files reviewed, 4 unresolved discussions (waiting on akash-a-n, gmarzot, michalhosna, mondain, Oxyd, peterchave, and TimEvens).

afrind:
What's left here was split between #180 and #181

PR #170, opened 2026-04-11
Pre-public hardening: drop App token from PR CI, add CODEOWNERS + SECURITY

gmarzot:
## Summary

Prepares moqx for public visibility by removing the only secret-bearing path from PR CI, adding governance files, and tightening macOS coverage.

### What changes

- **`scripts/setup-deps-tarball.sh`** — switch from authenticated workflow-artifact downloads (`gh api .../actions/workflows/.../runs` + `gh run download`, both of which require `actions:read` even on public repos) to **anonymous release-asset downloads via `curl`**. Uses moxygen's `snapshot-latest` GitHub release, parses the embedded commit SHA from the release body, and verifies it matches `deps/moxygen` HEAD. No `gh` CLI required, no auth required for read; works for fork PRs.
  - The api.github.com metadata fetch *is* authenticated when `GITHUB_TOKEN`/`GH_TOKEN` is set, to avoid the unauthenticated 60/hour/IP rate limit (observed as 403 on macOS-15 runners). Falls back to unauthenticated for local use.
  - The `Authorization` header is sent only to api.github.com — never to the release asset CDN, never echoed.
- **`.github/workflows/ci-pr.yml`** — drop `OMOQ_APP_ID` / `OMOQ_APP_PRIV_KEY` from the `build` job. Pass `secrets.GITHUB_TOKEN` (the default workflow token, NOT the App key) into the Setup dependencies step. After this PR merges, **PR CI runs without the App token**, which is the security model we need before flipping the repo public. The default `GITHUB_TOKEN` is read-only on fork PRs but still counts as authenticated for API rate-limit purposes.
- **`.github/workflows/ci-main.yml`** — add macOS to the push-to-main build matrix (mirrors the existing macOS coverage in ci-pr.yml) so post-merge runs also verify macOS. Many moqx developers work on macOS.
- **`scripts/build.sh`** — plumb a new `--use-latest` setup option through to `setup-deps-tarball.sh`. Also fixes an empty-array expansion under `set -u` (bash < 5.2).
- **`CODEOWNERS`** — auto-tags every prior reviewer of moqx PRs (`@afrind @akash-a-n @gmarzot @michalhosna @mondain @Oxyd @peterchave @suhasHere @TimEvens`) so they keep getting requested for reviews. The merge button is gated separately by branch protection (only `gmarzot` / `afrind` / `peterchave` + the sync bot, applied post-flip). Any approving review from a write-tier collaborator counts toward the gate; CODEOWNERS does not lock approval to the listed users.
- **`SECURITY.md`** — private vulnerability reporting via `security@openmoq.org`.

### SHA matching behavior

Today, `setup-deps-tarball.sh` finds the workflow run for the exact submodule SHA — works even for older SHAs (within the 90-day artifact retention window). After this PR, the snapshot-only fast path is strict-match by default:

| Submodule SHA | Behavior |
|---|---|
| Matches snapshot-latest | Uses snapshot ✓ |
| Older than snapshot-latest | Falls back to source build via `setup-deps-standalone.sh` (existing fallback in `build.sh`) |

The `moxygen-sync` workflow keeps `deps/moxygen` aligned with moxygen `main` continuously, so the strict-match case is the common case. Mismatches happen only when reviving an old branch or working off-tree.

### `--use-latest` and `--no-fallback`

Two complementary flags for `setup-deps-tarball.sh` (also exposed via `build.sh setup --use-latest`):

|  | (no flags) | `--no-fallback` | `--use-latest` |
|---|---|---|---|
| SHA matches | snapshot ✓ | snapshot ✓ | snapshot ✓ |
| SHA mismatch | falls back to source build | dies | uses snapshot anyway |
| Network failure | falls back to source build | dies | falls back to source build |

`--use-latest` is for "I want to track moxygen tip without bumping the submodule pin." Honors `MOQX_TARBALL_USE_LATEST=1` as an env var equivalent.

### What's NOT in this PR

- No publishing changes on moxygen — `snapshot-latest` is already public
- No branch-protection changes on moqx (cannot be set via API while private — applied immediately post-flip)
- No collaborator demotion — broad collaborator list stays as-is; gating is via the merge restrictions allowlist alone

## Test plan

- [ ] CI matrix (linux / macos / asan debug) passes after the App token is removed
- [ ] Manual local test: `unset GH_TOKEN; rm -rf .scratch; bash scripts/build.sh setup` succeeds against current submodule
- [ ] Manual local test: `bash scripts/build.sh setup --use-latest` proceeds when submodule != latest moxygen
- [ ] Confirm `setup-deps-tarball.sh` exits non-zero on intentional SHA mismatch and `build.sh` falls through to source build
- [ ] After merge: apply branch protection on moqx `main` and verify the `restrictions` allowlist locks merging to `afrind` + `gmarzot` + `peterchave` (+ `omoq-sync-bot` for sync auto-merge)

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/170)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,7 @@
+# CODEOWNERS — gates merging via branch protection.
+#
+# An approving review from at least one code owner is required for any PR
+# that touches this repository. The merge button itself is restricted (via
+# branch protection `restrictions`) to the same set of users.
+
+* @gmarzot @afrind
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,4 +4,4 @@
 # that touches this repository. The merge button itself is restricted (via
 # branch protection `restrictions`) to the same set of users.
 
-* @afrind @gmarzot @peterchave @suhasHere
+* @afrind @akash-a-n @gmarzot @michalhosna @mondain @Oxyd @peterchave @suhasHere @TimEvens
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,4 +4,4 @@
 # that touches this repository. The merge button itself is restricted (via
 # branch protection `restrictions`) to the same set of users.
 
-* @gmarzot @afrind
+* @afrind @gmarzot @peterchave
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,4 +4,4 @@
 # that touches this repository. The merge button itself is restricted (via
 # branch protection `restrictions`) to the same set of users.
 
-* @afrind @gmarzot @peterchave
+* @afrind @gmarzot @peterchave @suhasHere
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -47,6 +47,10 @@ jobs:
             preset: default
             build_dir: build
             runner: ubuntu-22.04
+          - name: macos
+            preset: default
+            build_dir: build
+            runner: macos-latest
           - name: asan debug
             preset: san
             build_dir: build-san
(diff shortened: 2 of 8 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,7 @@
+# CODEOWNERS — gates merging via branch protection.
+#
+# An approving review from at least one code owner is required for any PR
+# that touches this repository. The merge button itself is restricted (via
+# branch protection `restrictions`) to the same set of users.
+
+* @afrind @gmarzot @peterchave
```
suhasHere: I wonder if i can be added to the list of code owners for the time being.

gmarzot: this was provisional and was seeking input on who wants to be there.. adding now

PR #203, opened 2026-04-16
hotfix: defer system dep check until source build actually needed

gmarzot:
## Summary

`build.sh setup` checked system deps (libssl-dev, libglog-dev, etc.) eagerly — before even attempting the tarball download. This caused the publish job to fail on bare ubuntu-22.04 runners that only need to download a tarball, never compile.

### Fix

Move the dep check from before-tarball-attempt to right-before-source-build. The check now only runs when `mode == from-source` (either explicitly or after falling back from a failed tarball download). `from-release` paths never hit the check.

Also adds `--no-fallback` to ci-main's publish job setup call — publish only downloads a tarball for Docker image building, never compiles moxygen.

### Tested locally

```
--use-latest               → skips dep check, downloads tarball ✓
--use-latest --no-fallback → skips dep check, downloads tarball ✓
--from-source              → checks deps before source build ✓
```

## Test plan

- [x] CI passes (check-format, linux, asan debug)
- [ ] After merge: ci-main publish job succeeds on bare runner

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/203)

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -128,7 +128,7 @@ jobs:
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
           MOQX_PLATFORM: bookworm-amd64
-        run: bash scripts/build.sh setup --use-latest
+        run: bash scripts/build.sh setup --use-latest --no-fallback
 
       - name: Stage tarball for Docker build
         run: |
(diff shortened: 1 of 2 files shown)


afrind (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/203)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #218, opened 2026-04-17
fix: suppress peer namespace echo on reconnect

afrind:
When a peer relay reconnects and sends SUBSCRIBE_NAMESPACE, the relay was echoing back namespaces that originally came from that peer because the session identity check (sourceSession != newSession) passed for the new connection.

Tag each NamespaceNode with sourcePeerID and skip delivery when that ID matches the subscribing peer's relay ID, regardless of session identity. Also clear sourcePeerID in doPublishNamespaceDone alongside sourceSession so nodes never carry a stale peer tag after retraction.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/218)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -24,8 +24,12 @@ namespace openmoq::moqx {
 // no handle map needed.
 class MoqxRelayNamespaceHandle : public Publisher::NamespacePublishHandle {
 public:
-  MoqxRelayNamespaceHandle(std::weak_ptr<MoqxRelay> relay, std::shared_ptr<MoQSession> session)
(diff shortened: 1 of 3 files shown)


peterchave (approved):
:lgtm:

> @peterchave reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/218#-:-OqQfykg-XMNKgTaIrOh:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, gmarzot, michalhosna, mondain, Oxyd, suhasHere, and TimEvens).

PR #219, opened 2026-04-17
fix: clean up bridge-sourced namespaces on session close

afrind:
`MoqxRelayNamespaceHandle` now tracks the namespaces it has announced and calls `doPublishNamespaceDone` for each active one in its destructor.  In production the handle lives inside `SubNsStreamCallback` in a coroutine frame that is cancelled on ungraceful disconnect; without this fix those tree entries persisted with stale `sourceSession` shared_ptrs and downstream subscribers never received `NAMESPACE_DONE`.

Graceful `namespaceDoneMsg` removes entries from the tracking set so the destructor is a no-op on clean disconnect.  `doPublishNamespaceDone's` existing ownership guard `(sourceSession != session)` ensures the destructor is harmless if a new publisher has already taken over.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/219)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -7,6 +7,7 @@
  */
 
 #include "MoqxRelay.h"
+#include <folly/container/F14Set.h>
 #include <moxygen/MoQFilters.h>
 #include <moxygen/MoQTrackProperties.h>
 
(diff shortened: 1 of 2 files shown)


peterchave (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/219#-:-OqQga2f0Hr8ebHdqew8:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, gmarzot, michalhosna, mondain, Oxyd, suhasHere, and TimEvens).

PR #267, opened 2026-05-02
refactor: extract SubscriptionRegistry from MoqxRelay

afrind:
Encapsulates subscriptions_ behind a SubscriptionRegistry class with verb-centric APIs (getOrCreateFromSubscribe, createFromPublish, getUpstreamView, getFetchView, getTopNView, onPublisherTerminated, etc.) so that MoqxRelay no longer touches RelaySubscription fields directly.

Key design points:
- UpstreamSubscribePending RAII guards the window between entry creation and upstream subscribe completion; its destructor fires the failure path (promise.setException + erase) if complete() was never called
- getOrCreateFromSubscribe is synchronous on the first path and returns a Task<SubsequentSubscriber> on the subsequent path; the task is implemented as a proper coroutine function (not a lambda) so its parameters live in the heap-allocated coroutine frame rather than on the caller's stack
- Upstream-specific subReq overrides (locType=LargestObject, upstream priority/group order) are now applied after addSubscriber so the downstream subscriber's range reflects the client's original request

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/267)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,7 @@ FetchContent_MakeAvailable(reflectcpp)
 
 add_library(moqx_core STATIC
   src/NamespaceTree.cpp
+  src/SubscriptionRegistry.cpp
   src/MoqxRelay.cpp
   src/MoqxRelayContext.cpp
   src/MoqxRelayServer.cpp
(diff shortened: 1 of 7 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/267)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #289, opened 2026-05-05
ci: auto-delete sync-* branches when PRs closed without merge

gmarzot:
## Summary

Add a small workflow that auto-deletes the head branch when any PR with a `sync/*` or `sync-*` head ref is **closed without merging**. Fills the gap left by the repo-level "Automatically delete head branches" setting, which only fires on merge.

## Why

When a sync PR fails or is closed for retry, the branch lingers on remote. If the next sync run targets the **same upstream SHA** — rare but it happens when upstream hasn't advanced between attempts — the branch name collides and `git push` fails with non-fast-forward, blocking the cron cascade until the orphan is manually deleted.

Hit this exact failure mode today:
- picoquic: `sync/f54239a39145` collision after manual squash-merge of PR #7 left the branch on remote (`delete_branch_on_merge` fires only after `--merge`, not in all manual flows).
- moxygen: `sync-picoquic/0c516acd3b57` collision after manual close of PR #203 (no `--delete-branch`).

This workflow makes close-and-rerun reliable in all cases.

## Scope

- Triggers on `pull_request: closed`.
- Gated on `merged == false` (so it only runs for non-merge closes — `delete_branch_on_merge` already handles merges).
- Gated on head-ref prefix `sync/` or `sync-` — only automation-managed branches. Feature/fix/devops branches are unaffected.
- Skips fork PRs (we can't delete branches in someone else's fork).
- Uses default `GITHUB_TOKEN` with `contents: write` — no app token needed for a same-repo branch delete.

## Test plan

- [ ] After merge, manually close a sync-prefixed PR (existing or new) without merging — verify the branch disappears on remote.
- [ ] Manually close a non-sync PR (e.g., a feature/fix branch) — verify the branch is NOT deleted (workflow's `if:` guard rejects).
- [ ] Verify the workflow doesn't fire on PR `opened`, `synchronize`, or `merged` events (only `closed`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/289)

PR #291, opened 2026-05-05
ci: reclaim orphan sync branches inline in sync workflows

gmarzot:
## Summary

Add a defensive **orphan-reclaim** block to each sync workflow's create-branch step. When a sync PR is closed without merging, its head branch lingers on remote — the repo "Automatically delete head branches" setting only fires on merge, not close. If the next sync workflow run targets the same upstream SHA (rare but possible when upstream hasn't advanced between attempts), the branch name collides and `git push` fails with non-fast-forward, blocking the cron cascade.

## Behavior

Before each sync workflow's `git push`:

1. Check if the target branch exists on remote.
2. If yes, check whether any open PR uses it as head.
   - **Open PR exists** → abort with explicit error (defensive — PHASE A's blocking check should already have caught this).
   - **No open PR** → it's an orphan from a closed-without-merge attempt. Delete it.
3. Proceed with push as normal.

Net effect: closing a sync PR + re-running the workflow now works unconditionally, even when the upstream SHA hasn't changed between attempts. No manual branch cleanup, no separate cleanup workflow with skipped run noise.

## Why inline instead of a separate cleanup workflow

Earlier today we tried a `pull_request: closed` workflow (since-closed PR #10 in picoquic, closed PRs #205 in moxygen / #289 in moqx). It worked but produced "skipped" run records on every PR close in the repo (GitHub doesn't support head-branch filtering on `pull_request` triggers). Visual noise on the Actions UI made non-green entries unreliable as a problem signal.

Weaving the cleanup into the sync workflow itself fixes the same problem with no Actions UI noise — the cleanup only happens when actually needed, as part of the normal sync run.

## Test plan

- [ ] Manually close a sync PR without merging — confirm the orphan branch lingers (existing behavior).
- [ ] Re-run the sync workflow (`gh workflow run`) — confirm the orphan-reclaim warning fires and the new sync PR opens cleanly.
- [ ] Verify the abort path: if you somehow get into a state with an open PR for the target branch, the workflow should error explicitly rather than silently force-pushing.

## Related

- This PR removes `.github/workflows/sync-branch-cleanup-on-close.yml` — the close-trigger workflow merged via PR #10. Picoquic-only; the same change in moxygen / moqx never merged. (Picoquic PR only.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/291)

### DIFF
--- a/.github/workflows/moxygen-sync.yml
+++ b/.github/workflows/moxygen-sync.yml
@@ -149,6 +149,22 @@ jobs:
           TARGET_SHA="${{ steps.target.outputs.target_sha }}"
           SHORT_SHA="${{ steps.target.outputs.short_sha }}"
           BRANCH="sync-moxygen/${SHORT_SHA}"
+          REPO="${{ github.repository }}"
+
+          # Reclaim orphan branch if it exists with no open PR. PHASE A's
+          # blocking check rejects when an open sync-moxygen/* PR is pending,
+          # so by this point any pre-existing branch with no open PR is from a
+          # prior closed-without-merge sync attempt — safe to delete.
+          if git ls-remote --exit-code --heads origin "$BRANCH" >/dev/null 2>&1; then
+            EXISTING_PR=$(gh pr list --repo "$REPO" --state open --head "$BRANCH" \
+              --json number --jq '.[0].number // empty' 2>/dev/null || true)
+            if [ -n "$EXISTING_PR" ]; then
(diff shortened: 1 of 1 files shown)


afrind (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/291)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #302, opened 2026-05-08
MoqxCache: cap publisher cache duration, skip caching when mcd=0

afrind:
Adds setMaxAllowedCacheDuration() so operators can cap publisher-set cache durations. setTrackExtensions() now extracts the publisher's max cache duration extension, clamps it to the cap, and falls back to defaultMaxCacheDuration_ when the publisher doesn't set one. setMaxCacheDuration/clearMaxCacheDuration are now private; callers should go through setTrackExtensions.

Adds CacheTrack::shouldSkipCaching() which returns true when evicted or when maxCacheDuration is explicitly 0ms. defaultMaxCacheDuration=0ms acts as "opt-in only" mode: publisher must set a non-zero duration or objects are forwarded without being written to cache.

Wires cache.maxCacheDuration from CacheConfig into the new setMaxAllowedCacheDuration() call in MoqxRelay, removing the TODO.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/302)

### DIFF
--- a/src/MoqxCache.cpp
+++ b/src/MoqxCache.cpp
@@ -8,6 +8,7 @@
 
 #include "MoqxCache.h"
 #include <folly/logging/xlog.h>
+#include <moxygen/MoQTrackProperties.h>
 
 // Maxmimum cache size / per track? Number of groups
 // Fancy: handle streaming incomplete objects (forwarder?)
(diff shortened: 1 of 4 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/302)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #303, opened 2026-05-08
MoqxCache: fix two updateInProgress UB/crash bugs

afrind:
Branch 3 of updateInProgress() erases the fetchesInProgress entry and sets fetchInProgressIt_=end(). Two distinct bugs follow from this:

1. XCHECK crash (boundary collision): the previous re-key logic would attempt to re-key to maxLocation even when a concurrent fetch had already inserted an entry there. Fix: skip re-key when current >= maxLocation (erase-only), and guard the post-erase path so a second call from endOfFetch() doesn't attempt re-key on end().

2. UB dereference (double call): fetchInProgressIt_->second.progress was updated before the fetchInProgressIt_ != end() guard, so a second updateInProgress() call (e.g. endOfFetch() after a finFetch=true object) would dereference end(), corrupting the map sentinel. Fix: hoist the guard to cover the progress update, and add a matching guard to the else branch.

Add regression tests for both scenarios. The double-call test was verified to abort via a temporary XCHECK before the structural fix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/303)

### DIFF
--- a/src/MoqxCache.cpp
+++ b/src/MoqxCache.cpp
@@ -962,17 +962,26 @@ public:
     inProgress_.post();
     if (fetchRangeIt_.isValid()) {
       auto current = *fetchRangeIt_;
-      fetchInProgressIt_->second.progress = current;
-      // Re-key the map entry when crossing a group boundary so the
-      // old key slot is available for new fetches of evicted groups.
-      if (current.group > fetchInProgressIt_->first.group) {
-        auto entry = fetchInProgressIt_->second;
-        eraseAndMakeGroupEvictable();
-        emplaceAndPinGroup(current, entry);
+      // finFetch=true sets iterator to end(); guard prevents endOfFetch() re-entry dereference.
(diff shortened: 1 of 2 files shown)


suhasHere (approved):
> @suhasHere reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/303)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, gmarzot, michalhosna, mondain, Oxyd, peterchave, and TimEvens).

gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/303)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, and TimEvens).

PR #324, opened 2026-05-23
relay: adopt new forwardChanged(forwarder, bool) callback signature

afrind:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openmoq/moqx/pull/324).
* #325
* __->__ #324
* #323

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/324)

PR #325, opened 2026-05-23
stats: add keyed loop observer removal to fix use-after-free

afrind:
QuicStatsCollector::Callback registered an EVB loop observer capturing
data_.get() as a raw pointer.  If the Callback was destroyed while its
EventBaseStatsCollector was still alive, the observer would continue to
fire and dereference the freed pointer.

Fix by giving addLoopObserver a key (typically the owning object's `this`)
and adding removeLoopObserver(key), which posts an erase to the EVB thread.
Callback's destructor now calls removeLoopObserver so the subscription is
torn down safely.  PicoQuicStatsCollector's observer is similarly keyed
for consistency.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openmoq/moqx/pull/325).
* __->__ #325
* #324
* #323

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/325)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -979,7 +979,7 @@ void MoqxRelay::onEmpty(MoQForwarder* forwarder) {
   }
 }
 
-void MoqxRelay::forwardChanged(MoQForwarder* forwarder) {
+void MoqxRelay::forwardChanged(MoQForwarder* forwarder, bool forward) {
   const auto& ftn = forwarder->fullTrackName();
   auto upstreamView = registry_.getUpstreamView(ftn);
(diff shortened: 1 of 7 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/325)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #329, opened 2026-05-25
relay: add PublisherCrossExecFilter and SubscriberCrossExecFilter

afrind:
Summary:
Adds two executor-crossing filter classes that wrap the Publisher and
Subscriber interfaces and forward all calls to a target executor.

PublisherCrossExecFilter uses co_withExecutor on every coroutine method
(subscribe, fetch, trackStatus, subscribeNamespace) to switch to the
target executor before invoking the inner. goaway() is fire-and-forget
via executor->add().

SubscriberCrossExecFilter does the same for publishNamespace() and
goaway(). publish() is more involved: it returns a CrossExecFilter as
the consumer immediately (before the inner is called), then the reply
Task switches to the target executor, calls inner->publish(), and wires
setDownstream() to connect the real consumer. FIFO ordering on the
target executor guarantees setDownstream() runs before any data writes
enqueued afterward.

Both classes require the target executor to be FIFO if call ordering
matters. Unit tests cover success/error paths for all methods plus the
goaway enqueue-vs-invoke ordering.

Test Plan:
moqx_publisher_cross_exec_filter_test
moqx_subscriber_cross_exec_filter_test

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/329)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,8 @@ add_library(moqx_core STATIC
   src/relay/TopNFilter.cpp
   src/relay/PropertyRanking.cpp
   src/relay/CrossExecFilter.cpp
+  src/relay/PublisherCrossExecFilter.cpp
+  src/relay/SubscriberCrossExecFilter.cpp
 )
 
 target_include_directories(moqx_core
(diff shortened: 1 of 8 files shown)


gmarzot (approved):
> @gmarzot reviewed 8 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/329)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #333, opened 2026-05-25
chore(test): macOS dev hygiene — portable head + drop unused exec captures

gmarzot:
## Summary

Two small fixes that surface on a fresh Apple Silicon dev mac (e.g. the one being prepped as a moqx dev/test env) but not in CI, because GitHub's ``macos-15`` runner image happens to mask both.

## Fix 1: portable ``head`` in ``test/test_relay_chain.sh``

Line 129 was using ``head -n -1`` (GNU "all but last line"). BSD ``head`` on macOS rejects negative line counts:

```
head: illegal line count -- -1
```

Swapped to ``sed '$d'`` which means the same thing portably (drops the last line of input) on BSD ``sed`` and GNU ``sed`` alike.

This is the proximate cause of the ``relay_chain`` ctest going red on dev macs; the test is also known-flaky for other reasons, but at least the script no longer reliably fails before doing useful work.

## Fix 2: drop unused ``exec`` from two inner lambda captures in ``MoqxCacheTest.cpp``

clang emits two ``-Wunused-lambda-capture`` warnings for the inner ``.thenTry`` lambdas in ``TestFetchPartialMissTwoUpstreamFetchesTailDesc`` (line 1872) and ``TestFetchPartialMissTwoUpstreamFetchesTail`` (line 1898). The outer lambdas in those same expression chains genuinely use ``exec`` (via ``.via(exec)``) — those captures stay. Only the two inner captures are unused; this PR drops just those two.

All other ``[this, exec]`` captures in the file use ``exec`` somewhere in their body (``.via(exec)`` or ``exec->add(...)``) — verified by grep — so they are unchanged.

## Why CI didn't flag either before

- The ``head -n -1`` line only fires from a relay-chain test that's already known-flaky; CI's pass on it is partly luck.
- The lambda warnings are non-fatal; CI compiled with them and the build succeeded. They are just noise that confused dev-mac diagnostics when looking at the failed build's tail.

## Test plan

- [x] ``bash test/test_relay_chain.sh`` no longer fails with the BSD ``head`` error on macOS.
- [x] ``bash scripts/build.sh`` on macOS produces fewer warnings (the two -Wunused-lambda-capture diagnostics from this file are gone).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/333)

### DIFF
--- a/test/MoqxCacheTest.cpp
+++ b/test/MoqxCacheTest.cpp
@@ -1869,7 +1869,7 @@ CO_TEST_F(MoqxCacheTest, TestFetchPartialMissTwoUpstreamFetchesTailDesc) {
 
         expectUpstreamFetch({1, 3}, {3, 0}, 0, AbsoluteLocation{3, 0}, GroupOrder::NewestFirst)
             .via(exec)
-            .thenTry([this, exec](const auto&) {
+            .thenTry([this](const auto&) {
               expectFetchObjectsDescending({1, 3}, {4, 0}, true, 10, 1, 1, true);
               serveCacheRangeFromUpstreamDescending({1, 3}, {4, 0}, 10, 1, 1, true);
             });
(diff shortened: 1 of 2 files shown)


afrind (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/333)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #350, opened 2026-05-28
bpf: attach classic reuseport steering filter to QUIC worker sockets

afrind:
Adds MOQX_ENABLE_BPF_STEERING (default ON on Linux) which overrides mvfst's weak mvfst_hook_on_socket_create hook to attach a cBPF filter that routes non-initial QUIC packets directly to the owning worker by decoding the workerId from mvfst's default V1 connection ID encoding. Long-header (initial/handshake) packets are spread by UDP source port.

Note: threads is still forced to 1 so this doesn't do anything _yet_ but it will real soon now (TM)

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/350)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,15 @@ option(MOQX_BUILD_TESTS "Build tests" ON)
 option(MOQX_BUILD_BENCHMARKS "Build benchmarks" OFF)
 option(MOQX_ENABLE_SANITIZERS "Enable ASAN/UBSAN (non-Release)" OFF)
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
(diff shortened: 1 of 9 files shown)


afrind:
The setting is global (across all mvfst listeners), so I put it in the global section and gave it a mvfst prefix.  If we add another global setting, we can refactor to mvfst: global section.

gmarzot (approved):
> @gmarzot reviewed 9 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/350)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #352, opened 2026-05-28
relay: wrap consumer in CrossExecFilter inside PublisherCrossExecFilter

afrind:
subscribe() and fetch() now capture the caller's executor and wrap the consumer before dispatching to targetExec_, so upstream data is delivered back on the caller's executor without call-site boilerplate. On fetch error, inner_ never stored the consumer so no lambdas are in-flight; deactivate() breaks the selfGuard_ cycle to prevent a leak.

Test: FetchErrorDoesNotCallEndOfFetch asserts endOfFetch is not called on error; FetchReturnsSuccess is updated to run on an EventBase so the captured callerExec stays alive for the endOfFetch drain.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/352)

### DIFF
--- a/src/relay/PublisherCrossExecFilter.cpp
+++ b/src/relay/PublisherCrossExecFilter.cpp
@@ -5,6 +5,7 @@
  */
 
 #include "relay/PublisherCrossExecFilter.h"
+#include "relay/CrossExecFilter.h"
 
 namespace openmoq::moqx {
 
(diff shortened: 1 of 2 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/352)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #367, opened 2026-06-01
perf-test: increase packets read per loop to 256

afrind:
---
[//]: # (BEGIN SAPLING FOOTER)
* #368
* __->__ #367
* #366
* #365
* #364
* #363
* #362
* #361
* #360

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/367)

PR #366, opened 2026-06-01
config: add udp_socket_buffer_bytes listener config option

afrind:
Exposes udpSendBufferBytes/udpRecvBufferBytes via MoQServer::Options,
plumbed through ParsedConfig → ConfigResolver → MoqxRelayServer.
0 (default) keeps the existing MoQServer built-in 1 MB behaviour.
Also comments out reorderingThreshold (removed from moxygen API).

---
[//]: # (BEGIN SAPLING FOOTER)
* #368
* #367
* __->__ #366
* #365
* #364
* #363
* #362
* #361
* #360

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/366)

PR #365, opened 2026-06-01
relay: per-thread local forwarder data path (use_local_forwarders)

afrind:
Adds an optional data-plane mode in which each I/O thread owns its own
MoQForwarder per track, so fanning objects out to subscribers on the same
thread needs no cross-thread hop. Gated by the new use_local_forwarders
option (which requires use_relay_thread); off by default.

Theory of operation
--------------------
The relay now runs in one of three modes, chosen per session by
createPublisherFilter()/createSubscriberFilter():

1. Single-threaded (relayExec_ unset): relay state and all forwarders live
   on the session I/O thread; MoqxRelay is the session handler directly.

2. Cross-exec (relayExec_ set, local forwarders off): relay state is
   isolated on a dedicated relay executor. Session handlers are
   Publisher/SubscriberCrossExecFilter, which hop to the relay exec before
   touching state. One primary forwarder per track lives on the relay exec
   and fans out to every subscriber via a per-callback cross-exec hop.

3. Local forwarders (relayExec_ set, local forwarders on): the primary
   forwarder for a track lives on its publisher's I/O thread. Each
   subscriber I/O thread keeps a thread-local forwarder (LocalForwarderRegistry
   in tlForwarders_) that attaches to the primary as a single channel
   subscriber. The primary thus hops once per subscriber *thread*; each local
   forwarder then fans out to its same-thread subscribers with no further hop.

Subscribe (local mode) is split so the relay-exec critical section stays
small: subscribeFromSubscriberThread orchestrates on the subscriber thread,
dispatching subscribeStatefulWork (registry lookup, first-vs-subsequent
decision) to the relay exec, then wiring the local->primary channel; for the
first subscriber, completeFirstSubscriber issues the upstream subscribe and
applies the SubscribeOk on the primary's exec. Publish (local mode) makes the
publisher's local forwarder the primary and registers it on the relay exec via
setupPublisherPrimary; the relay chain (TopNFilter -> termination -> cache)
attaches to it as a passive channel subscriber so it observes all objects
without counting as a forwarding subscriber.

Lifecycle callbacks flow LocalForwarderCallback -> CrossExecForwarderCallback
-> ChannelForwarderCallback (on the primary's exec), which propagates forward
and new-group-request changes upstream via requestUpdate and removes the
channel subscriber on onEmpty. LocalForwarderRegistry removal is
identity-checked so teardown is order-independent. PendingForwarderCallback
captures forwardChanged/newGroupRequested/onEmpty during the setup window for
[...]

PR #363, opened 2026-06-01
test: add MultiThread relay test mode

afrind:
Add RelayMode::MultiThread to the MoQRelayTest parameterized suite so the
relay is exercised across an executor boundary, alongside the existing
SingleThread mode.

Fixture (MoqxRelayTestFixture.h/.cpp, MoqxRelayTestModes.cpp):
- New RelayMode::MultiThread enum value, instantiated in the AllModes suite.
- SetUp() starts a ScopedEventBaseThread, wires the relay onto it via
  setRelayExec, and wraps it in PublisherCrossExecFilter /
  SubscriberCrossExecFilter. TearDown() drains pending relay-exec tasks and
  tears down in order to avoid use-after-free on NamespaceTree.
- TestMoQExecutor::drive() does a loopOnce(), then in MT mode round-trips
  through the relay EVB twice to flush task->relay and relay-created tasks.
- Add driveIfMultiThread(), make verifyOnRelayExec() hop to the relay EVB,
  and add driveUntil(pred) to advance async cascades deterministically
  instead of using a fixed drive count (capped at one iter in ST mode).

Tests (DataPlane, NGR, Peer, Publish, SubNs, Subscribe):
- Insert driveIfMultiThread() where work crosses executors.
- Replace fixed-drive-then-read with driveUntil() plus std::atomic flags.
- DuplicateSubgroupCancelledWhenNoActiveConsumers splits ST/MT expectations:
  in MT the CrossExecFilter defers the CANCELLED error to the next op, so it
  probes via endOfSubgroup().
- Add reset() calls to simulate the publisher/QUIC resetting open streams.

Test Plan:

---
[//]: # (BEGIN SAPLING FOOTER)
* #365
* #364
* __->__ #363
* #362
* #361

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/363)

PR #368, opened 2026-06-01
relay: make stop() idempotent and keep context_ alive during teardown

afrind:
Guard stop() with a stopped_ flag instead of context_ presence, and stop
resetting context_. terminateClientSession can run after stop() returns,
from handleClientSession coroutines still draining on the evb/IO threads,
so context_ must outlive stop().

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

---
[//]: # (BEGIN SAPLING FOOTER)
* __->__ #368
* #367
* #366
* #365
* #364
* #363
* #362
* #361
* #360

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/368)


afrind:
Replacing with #372

PR #371, opened 2026-06-01
bpf: attach reuseport steering filter exactly once per group

afrind:
SO_ATTACH_REUSEPORT_CBPF is a property of the kernel reuseport group, not the individual fd: re-attaching replaces and frees the group's single prog while sibling sockets' RX softirqs may still be running it. The mvfst_hook_on_socket_create hook fired on every invocation — once per worker listener fd at bind() and again per accepted connection at runtime (site 2 wraps the worker's same listener fd) — turning that into a hot, multi-threaded use-after-free.

Dedup attachment per reuseport group, keyed by the bound address:port (via getsockname). State is encapsulated in a ReuseportSteering class held in a folly::Indestructible singleton (never destructs, so a late hook call from a draining IO thread can't touch freed state on shutdown):

- A mutex-guarded F14 set of already-attached SocketAddresses. The lock is held across check -> setsockopt -> insert so concurrent workers binding the same group serialize and exactly one attaches.
- A per-thread folly::ThreadLocal F14 set of resolved fds as a lock-free, syscall-free fast path for the steady-state per-connection case.

getsockname/setsockopt failures return without caching so a later hook retries; client sockets (where the option legitimately fails) re-run the rate-limited slow path harmlessly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/371)

### DIFF
--- a/src/bpf/QuicReuseportSteering.cpp
+++ b/src/bpf/QuicReuseportSteering.cpp
@@ -23,18 +23,54 @@
 #include <atomic>
 #include <cerrno>
 #include <cstring>
+#include <mutex>
 
 #include <linux/filter.h>
 #include <sys/socket.h>
 
+#include <folly/Indestructible.h>
+#include <folly/SocketAddress.h>
+#include <folly/ThreadLocal.h>
+#include <folly/container/F14Set.h>
 #include <folly/logging/xlog.h>
+#include <folly/net/NetworkSocket.h>
 
 namespace openmoq::moqx {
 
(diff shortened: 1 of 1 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/371)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #376, opened 2026-06-02
Fix Debian bookworm builds and switch deps to CPM

TimEvens:
- Switch yaml-cpp and reflect-cpp from CMake FetchContent to CPM for cleaner dependency management and fewer CMake version workarounds.
- Harden the Debian bookworm native and docker builds. No longer require pip install of cmake. 
- Use Boost static libs when building standalone moxygen deps.
- Add CLion setup instructions to BUILD.md

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/376)

PR #385, opened 2026-06-04
setup-deps-standalone: prefer static Boost when available, with env override

gmarzot:
## Summary
#376 added `-DBoost_USE_STATIC_LIBS=ON` unconditionally, which fails the standalone moxygen dep build on distros that don't ship `libboost_*.a` (CentOS / RHEL — reported by devs using those systems). This restores the build on those distros while keeping the *"static when possible"* policy from #376.

## How it picks

The probe uses `cmake --find-package` so it consults CMake's own `find_package(Boost)` algorithm — no hardcoded library paths, no distro sniffing, no `pkg-config` (Boost's `.pc` files are inconsistent across distros). Same code path as the real configure step, so the answer is definitive.

| `BOOST_USE_STATIC_LIBS` env | behavior |
|---|---|
| `auto` (default — including unset/empty) | probe via `cmake --find-package`; pass the flag if static is findable |
| `on` / `ON` / `1` / `true` | force static — `cmake` configure fails loudly if static Boost is missing (useful in publish CI to catch env regressions) |
| `off` / `OFF` / `0` / `false` | force shared (omit the flag) |
| anything else | error + exit 1 with a usage hint |

A one-line `==> Boost linking: …` log makes the chosen mode visible in build output.

## Test plan
- [x] Smoke-tested each mode locally (unset, auto, on, off, empty, bogus) — all behave per spec.
- [ ] Verify on a CentOS / RHEL dev box that `auto` now selects shared and the build succeeds (was failing on `=ON`).
- [ ] Verify on Debian/Ubuntu that `auto` still selects static (no behavior change from #376).
- [ ] Spot-check `BOOST_USE_STATIC_LIBS=on` and `=off` produce the expected configure args.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/385)

### DIFF
--- a/scripts/setup-deps-standalone.sh
+++ b/scripts/setup-deps-standalone.sh
@@ -97,6 +97,35 @@ if [[ -n "$CXX_FLAGS" ]]; then
   EXTRA_CMAKE_ARGS+=("-DCMAKE_CXX_FLAGS=${CXX_FLAGS}")
 fi
 
+# Boost linking: prefer static when available (more portable artifacts);
+# fall back to shared on distros that don't ship libboost_*.a (CentOS/RHEL).
+# Override via env: BOOST_USE_STATIC_LIBS={auto,on,off}
+#   auto (default) — ask cmake's own find_package whether static is findable
+#   on             — force static (cmake configure fails loudly if missing)
+#   off            — force shared (omit the flag entirely)
+BOOST_STATIC_ARG=()
+case "${BOOST_USE_STATIC_LIBS:-auto}" in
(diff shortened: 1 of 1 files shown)


afrind (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/385)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #388, opened 2026-06-05
Test CI PR workflow - for fix/auto regression test

peterchave:
Testing the perf test is triggered on PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/388)

PR #389, opened 2026-06-05
setup-deps-standalone: probe compiler for static Boost, not cmake --find-package

afrind:
The auto Boost-linking probe from #388 used
`cmake --find-package ... -DMODE=EXIST -DBoost_USE_STATIC_LIBS=ON`, which is a false positive: that mode only checks whether Boost is locatable at all and ignores Boost_USE_STATIC_LIBS. On distros that ship only libboost_*.so (CentOS/RHEL), it reports "static found", the script selects static linking, and the build dies at ninja time with `libboost_context.a ... missing` -- the exact failure #388 meant to fix.

Replace the probe with one that asks the compiler driver to resolve each static archive moxygen actually links (context, filesystem, program_options, regex) via `c++ -print-file-name`. That returns an absolute path only when the .a exists in the compiler's own search paths -- definitive, no hardcoded paths, no distro sniffing.

Also always pass -DBoost_USE_STATIC_LIBS explicitly (ON or OFF) instead of omitting it for shared. A re-run of `setup` reuses the existing build dir; omitting would leave a stale Boost_USE_STATIC_LIBS=ON in CMakeCache.txt in force.

Verified: clean build picks shared and links; a poisoned =ON cache is overridden to OFF and still builds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/389)

### DIFF
--- a/scripts/setup-deps-standalone.sh
+++ b/scripts/setup-deps-standalone.sh
@@ -100,31 +100,46 @@ fi
 # Boost linking: prefer static when available (more portable artifacts);
 # fall back to shared on distros that don't ship libboost_*.a (CentOS/RHEL).
 # Override via env: BOOST_USE_STATIC_LIBS={auto,on,off}
-#   auto (default) — ask cmake's own find_package whether static is findable
+#   auto (default) — probe the compiler for the static archives moxygen needs
 #   on             — force static (cmake configure fails loudly if missing)
-#   off            — force shared (omit the flag entirely)
-BOOST_STATIC_ARG=()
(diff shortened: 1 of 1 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/389)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n, michalhosna, mondain, Oxyd, peterchave, suhasHere, and TimEvens).

PR #406, opened 2026-06-15
relay: gate bidi NAMESPACE forward on negotiated draft >=16

afrind:
doPublishNamespace chose the draft-16+ bidi NAMESPACE path versus the draft <=15 separate-stream PUBLISH_NAMESPACE path by testing whether the publish handle was present. The relay session populates that handle regardless of negotiated version, so a draft <=15 subscriber was routed into NamespacePublishHandle::namespaceMsg, which reaches an unimplemented moxygen SubNSReply stub (XLOG(FATAL)) and aborts the relay. Gate on the negotiated version instead, matching doPublishNamespaceDone.

Extend the existing SubNs tests to pass a non-null publish handle (as the relay session always does): ExactNamespaceSubscriberReceivesPublishNamespace now asserts a pre-v16 subscriber forwards via the separate stream and never the bidi handle (failing against the un-gated code), and EmptyPrefixAllowedV16 asserts a v16 subscriber forwards via the bidi handle.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/406)

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -12,7 +12,7 @@ listeners:
       # key_file: /path/to/key.pem     # Required when insecure: false
       insecure: true          # Skip TLS for local development
     endpoint: "/moq-relay"    # WebTransport endpoint path
-    # moqt_versions: [14, 16] # MOQT draft versions (empty = all supported)
+    # moqt_versions: [14, 16] # MOQT draft versions (empty = default 14,16)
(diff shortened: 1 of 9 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/406)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #407, opened 2026-06-15
test/relay_chain: re-enable Direction 4 gating and macOS run

afrind:
The relay no longer aborts on the --publish push path, so Direction 4 delivers data: restore it to a hard failure instead of XFAIL. Also drop the macOS deterministic skip so the test runs on Darwin again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/407)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/407)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #421, opened 2026-06-20
feat: add mvfst ignore_path_mtu config option

afrind:
Expose mvfst's canIgnorePathMTU as an mvfst.ignore_path_mtu config field, defaulting to false so packets stay at the conservative default size. When enabled, mvfst sends full max-UDP-payload packets, which improves throughput on networks where the path MTU is known to be large.

Enable it in perf-test.sh, where larger packets give a more representative best-case throughput number. This is a benchmarking knob and is left off by default for production safety.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/421)

### DIFF
--- a/scripts/perf-test.sh
+++ b/scripts/perf-test.sh
@@ -223,6 +223,7 @@ listeners:
     endpoint: "$ENDPOINT"
     mvfst:
       enable_gso: true
+      ignore_path_mtu: true
       max_conn_packets_sent_per_loop: 16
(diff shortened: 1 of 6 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/421)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #362, opened 2026-06-01
relay: isolate relay state on dedicated executor to support multiple I/O threads

afrind:
---
[//]: # (BEGIN SAPLING FOOTER)
* #365
* #364
* #363
* __->__ #362
* #361

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/362)

PR #424, opened 2026-06-20
 Run upstream-handle ops on the publisher exec; pin integration tests to draft 16

afrind:
Our integration tests are negotiating 14, because our alpn lists it first.  When testing 18, I discovered an MT bug from #362 / #365.  One commit here fixes the bug and the other pins the integration tests to 16 (which also demonstrated the bug), until 18 is ready.

Fixes a multi-threaded relay bug: `forwardChanged`, `newGroupRequested`, and
`onEmpty` operate on the upstream subscription handle (`requestUpdate` /
`unsubscribe`), which mutate the publisher's session inline without self-hopping,
so they must run on the publisher's executor, not the relay exec. Also covers the
publish-reconnect eviction path, which had the same latent bug against the old
publisher's session. No-op in single-threaded mode.

I considered wrapping the subscription handle so e.g. ->unsubscribe automatically jumps to the correct exec.  This was a little messier, so decided to go with a more bespoke solution for now.

Also pins the shell integration-test relays to a shared `MOQT_TEST_VERSIONS`
(`[16]`) so they exercise the chosen draft instead of silently negotiating down;
`test_conformance.sh` is left parameterized to keep running against all drafts.

Full ASan suite passes 655/655.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/424)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -593,7 +593,10 @@ MoqxRelay::PublishSetupResult MoqxRelay::publishWithSession(
     auto& evicted = *publishEntry.evicted;
     // Null handle => previous publisher already terminated and onPublishDone() tore it down; skip.
     if (evicted.handle) {
-      evicted.handle->unsubscribe();
+      // unsubscribe mutates the old publisher's session inline, so hop to its exec.
+      runOnSessionExec(relayExec_, evicted.publisherExec, [h = evicted.handle] {
+        h->unsubscribe();
+      });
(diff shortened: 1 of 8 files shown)


gmarzot (approved):
> @gmarzot reviewed 8 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/424)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #431, opened 2026-06-23
tools: moq_decode.py — decode d18 SUBSCRIBER_PRIORITY/GROUP_ORDER/FORWARD as uint8 (§10.2.7/8/12)

### LINKED ISSUE #430: moq_decode.py: d18 SUBSCRIBER_PRIORITY/GROUP_ORDER/FORWARD param values decoded as varint, should be uint8 (§10.2.7/8/12)
## Summary

`tools/moq_decode.py` decodes the **SUBSCRIBER_PRIORITY (0x20)**, **GROUP_ORDER (0x22)**, and **FORWARD (0x10)** parameter values as varints at draft-18, but §10.2.7 / §10.2.8 / §10.2.12 define each of these as a **uint8**. The draft permits this per-Type override of the generic §1.4.3 even-key=varint rule ("the serialization defined by that Type").

Because a varint and a uint8 are identical for values 0–63, the bug is silent until a value is ≥64 — most notably the **default `subscriber_priority` of 128**, which always trips it.

## Repro

A draft-18 Standalone FETCH with `subscriber_priority=128`, `group_order=1`:

```
python tools/moq_decode.py --version 18 16 00 23 01 01 02 08 6d 6f 71 2d 74 65 73 74 07 69 6e 74 65 72 6f 70 05 66 65 74 63 68 00 00 00 00 02 20 80 02 01
```

The decoder reads the uint8 `0x80` (=128) as the first byte of a 2-byte varint, consumes the following `0x02`, and then mis-aligns the rest of the message, ending in `! Error at offset 0x26: underflow reading MoQ varint`. moxygen's `MoQFramer` (`paramEncodingV18` → `ParamValueEncoding::Uint8`) parses the same bytes cleanly.

## Scope

Only the standalone `moq_decode.py` annotator is affected — the moxygen-based relay parses these correctly (this is a tool-only divergence).

## Suggested fix

In `p_params`, add a per-key special-case mirroring the existing `LARGEST_OBJECT` / `SUBSCRIPTION_FILTER` handling: for `draft >= 18` and `abs_key in (0x10, 0x20, 0x22)`, read the value with `read_uint8()` instead of `read_varint()`.

gmarzot:
`p_params` decoded the SUBSCRIBER_PRIORITY (0x20), GROUP_ORDER (0x22), and FORWARD (0x10) parameter values as varints (the generic §1.4.3 even-key rule). draft-18 §10.2.7 / §10.2.8 / §10.2.12 define each as a **uint8**, which §1.4.3 permits as a per-Type serialization override. The two forms are identical for values 0–63, so the bug was silent until a value ≥64 — most commonly the default `subscriber_priority=128`, which made the decoder read the uint8 `0x80` as the start of a 2-byte varint and mis-align the rest of the message.

This adds a `draft >= 18 and abs_key in (0x10, 0x20, 0x22)` branch using `read_uint8()`, placed before the generic even-key varint case — mirroring the existing per-key special-cases (`LARGEST_OBJECT`, `SUBSCRIPTION_FILTER`) and the file's `draft >=` version-gating. Drafts < 18 are unchanged (these stay varint there).

**Verification** — a draft-18 Standalone FETCH with `subscriber_priority=128`:
```
python tools/moq_decode.py --version 18 16 00 23 01 01 02 08 6d 6f 71 2d 74 65 73 74 07 69 6e 74 65 72 6f 70 05 66 65 74 63 68 00 00 00 00 02 20 80 02 01
```
Before: `! Error at offset 0x26: underflow reading MoQ varint`. After: `param[0].value 128`, `param[1].value 1`, clean — matching moxygen's `MoQFramer` (`paramEncodingV18 → Uint8`).

Fixes #430

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/431)

### DIFF
--- a/tools/moq_decode.py
+++ b/tools/moq_decode.py
@@ -410,6 +410,12 @@ def p_params(cursor, annot, draft, count, param_keys):
             blob_start = cursor.pos
             raw = cursor.read_bytes(vlen)
             p_subscription_filter_blob(raw, blob_start, annot, i, draft)
+        elif draft >= 18 and abs_key in (0x10, 0x20, 0x22):
+            # d18 §10.2.7/8/12: FORWARD (0x10), SUBSCRIBER_PRIORITY (0x20) and
+            # GROUP_ORDER (0x22) values are a fixed uint8, overriding the
+            # generic even-key varint (§1.4.3). Diverges at value >= 64.
+            val, vs = cursor.read_uint8()
+            annot.add(vs, cursor.pos, f"param[{i}].value", str(val))
         elif abs_key % 2 == 0:
             # Even-keyed params are plain varints
             val, vs = cursor.read_varint()


afrind (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/431)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gmarzot).

PR #434, opened 2026-06-23
fix: clean error instead of SIGABRT on unreadable TLS cert (#173)

### LINKED ISSUE #173: Crash on unreadable TLS certificate (SIGABRT instead of clean error)
## Description

moqx crashes with `SIGABRT` when it cannot read the TLS certificate/key files (e.g., permission denied on Let's Encrypt private key). Instead of a graceful error message and clean exit, the process dumps core.

## Reproduction

```bash
# Run as non-root user with root-only cert files:
/path/to/moqx --config /tmp/moqx.yaml
# where config references /etc/letsencrypt/live/.../privkey.pem (mode 0600, owner root)
```

## Observed behavior

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  no certificates read
*** Signal 6 (SIGABRT) received ***
Aborted (core dumped)
```

Stack trace shows the unhandled throw originates in fizz's `CertUtils.cpp:150` (`selfCertFromDataInternal`), called from `MoqxRelayServer.cpp:49` (`buildFizzContext`).

## Expected behavior

moqx should catch the exception during startup, print a clear error like:

```
Error: failed to load TLS certificate — check file paths and permissions
  cert: /etc/letsencrypt/live/.../fullchain.pem
  key:  /etc/letsencrypt/live/.../privkey.pem
```

and exit with a non-zero status code (no core dump).

## Environment

- moqx commit: e55d06d2
- moxygen: 5ae4061
- OS: Linux 6.6.87 (WSL2)

gmarzot:
Closes #173.

## Problem
When the configured TLS cert/key cannot be read (missing file, or root-only Let's Encrypt key read by a non-root process), `createFizzServerContext` throws deep in `fizz` ("no certificates read"). The exception escaped through the `MoqxRelayServer` constructor to `std::terminate` → **SIGABRT / core dump**, with no indication of which path was at fault.

## Fix
- **`buildFizzContext`** ([src/MoqxRelayServer.cpp](src/MoqxRelayServer.cpp)): wrap the real-TLS `createFizzServerContext` call and rethrow a `std::runtime_error` enriched with the offending cert/key paths.
- **`main`** ([src/main.cpp](src/main.cpp)): wrap listener construction in try/catch; on failure print a clear `Error: …` and `return 1`. Deliberately a clean non-zero exit (not `XLOG(FATAL)`, which would `abort()`/core-dump).

## Verified locally
Incremental build is clean. Running the binary with a nonexistent cert (`insecure: false`):

```
Error: failed to initialize listener: failed to load TLS certificate/key
(cert='/tmp/nope-cert.pem', key='/tmp/nope-key.pem'): no certificates read
- check the paths exist and are readable by this process
```

Exit code **1** (was 134 / SIGABRT before). No core dump.

## Scope
The catch in `main` covers all listener-setup failures, not just TLS, so other startup config errors also exit cleanly instead of terminating.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/434)

### DIFF
--- a/src/MoqxRelayServer.cpp
+++ b/src/MoqxRelayServer.cpp
@@ -42,12 +42,24 @@ buildFizzContext(const config::ListenerConfig& cfg) {
               ""
           );
         } else {
-          return quic::samples::createFizzServerContext(
-              alpns,
-              fizz::server::ClientAuthMode::Optional,
-              tls.certFile,
-              tls.keyFile
-          );
+          // createFizzServerContext throws (deep in fizz) when the cert/key
+          // can't be read or contain no certificate. Enrich the message with
+          // the offending paths so the caller can report it cleanly instead of
+          // letting an opaque "no certificates read" escape to std::terminate.
+          try {
+            return quic::samples::createFizzServerContext(
+                alpns,
+                fizz::server::ClientAuthMode::Optional,
+                tls.certFile,
+                tls.keyFile
+            );
+          } catch (const std::exception& e) {
+            throw std::runtime_error(
+                "failed to load TLS certificate/key (cert='" + tls.certFile + "', key='" +
+                tls.keyFile + "'): " + e.what() +
+                " - check the paths exist and are readable by this process"
+            );
+          }
         }
       },
       cfg.tlsMode
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -128,8 +128,15 @@ int main(int argc, char* argv[]) {
   auto statsRegistry = std::make_shared<stats::StatsRegistry>();
 
   std::vector<std::shared_ptr<moxygen::MoQServerBase>> servers;
-  for (const auto& listenerCfg : config.listeners) {
-    servers.emplace_back(makeRelayServer(listenerCfg, context, ioExecutor.get(), statsRegistry));
+  try {
+    for (const auto& listenerCfg : config.listeners) {
+      servers.emplace_back(makeRelayServer(listenerCfg, context, ioExecutor.get(), statsRegistry));
+    }
+  } catch (const std::exception& e) {
+    // Listener setup (e.g. TLS cert loading) can throw. Report cleanly and exit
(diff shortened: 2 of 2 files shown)


afrind (approved):
> @afrind reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/434)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gmarzot).

___
*[`src/MoqxRelayServer.cpp` line 47 at r1](https://reviewable.io/reviews/openmoq/moqx/434#-Ovu9acC9rlQkr_1IjZW:-Ovu9acC9rlQkr_1IjZX:b-wy1s42) ([raw file](https://github.com/openmoq/moqx/blob/0e0e83670210e49b55291451f79d7d85e0b138ae/src/MoqxRelayServer.cpp#L47)):*
> ```cpp
>           // createFizzServerContext throws (deep in fizz) when the cert/key
>           // can't be read or contain no certificate. Enrich the message with
>           // the offending paths so the caller can report it cleanly instead of
> ```

I would end the comment after "cleanly"
___
*[`src/main.cpp` line 138 at r1](https://reviewable.io/reviews/openmoq/moqx/434#-Ovu9hgR7EL2V0Ww5_8o:-Ovu9hgR7EL2V0Ww5_8p:bfzyy4x) ([raw file](https://github.com/openmoq/moqx/blob/0e0e83670210e49b55291451f79d7d85e0b138ae/src/main.cpp#L138)):*
> ```cpp
>     // Listener setup (e.g. TLS cert loading) can throw. Report cleanly and exit
>     // non-zero rather than letting it reach std::terminate / SIGABRT (#173).
>     std::cerr << "Error: failed to initialize listener: " << e.what() << "\n";
> ```

Is XLOG(ERR) more idiomatic?

Also, failing to start because one listener cert is unreadable may not be the best all the time.  Another option is to skip that listener and bring up the working ones.  This is fine for now, but could be a future config option.

PR #443, opened 2026-06-24
relay: don't inspect the publisher forwarder from relayExec_ on publishDone (LF mode)

afrind:
In LocalForwarder mode the publisher's forwarder is owned by the publisher's executor ([Pub]), but onPublishDone runs on relayExec_ via the passive relay chain. Its call to registry_.onPublisherTerminated() did rsub.forwarder->empty(), reading the [Pub]-owned forwarder's subscribers_ map from relayExec_ while the publisher mutated it — a data race (TSAN-confirmed; intermittent ASan heap-buffer-overflow).

The registry entry in LF mode only indexes a live publisher for subscriber discovery; existing subscribers are served off-relay via channel subs and the relay chain is held by the passive pigtail, so neither depends on the entry. Drop the entry outright in LF mode without touching the forwarder. Non-LF modes keep the empty()-conditional erase (single-exec, safe).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/443)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -408,10 +408,16 @@ void MoqxRelay::onPublishDone(const FullTrackName& ftn) {
     namespaceTree_.unpublishTrack(ftn.trackNamespace, ftn.trackName);
   }
 
-  // Clears handle + upstream; erases if no subscribers remain.
-  auto forwarder = registry_.onPublisherTerminated(ftn);
-  if (!forwarder) {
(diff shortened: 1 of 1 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/443)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #444, opened 2026-06-24
stats: per-thread MoQStatsCollector for MT-safe collection

afrind:
The stats collector predated multi-threading: one shared collector was wired to every session, so its plain counters/histograms were mutated concurrently from all io threads (TSAN-confirmed races in onSubscriptionEnd/onSessionEnd/etc.).

Pre-create one collector per io thread (initThreadStatsCollectors, called once at startup before serving), bound to its thread via a ThreadLocal. Each session wires its callbacks to its own thread's collector, so every mutation stays single-threaded. The StatsRegistry already aggregates across collectors for /metrics, so per-thread snapshots sum correctly.

/state's active-session read is stubbed to 0 for now (print-only, not asserted; deferred to the /state rework).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/444)

### DIFF
--- a/src/MoqxRelayContext.cpp
+++ b/src/MoqxRelayContext.cpp
@@ -173,25 +173,32 @@ folly::coro::Task<size_t> MoqxRelayContext::purgeCache(
   co_return total;
 }
 
+void MoqxRelayContext::initThreadStatsCollectors(folly::IOThreadPoolExecutor& ioExecutor) {
+  if (!statsRegistry_) {
+    return;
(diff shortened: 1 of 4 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/444)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #449, opened 2026-06-25
relay: hop cross-exec handle teardown to the owning exec

afrind:
The reciprocal peer SubscribeNamespaceHandle was acquired via a direct session->subscribeNamespace() and stored raw, so peerSubNsHandles_.erase() destroyed it on relayExec_ — running its teardown (isClosed() read + control RST) on the wrong thread while the peer session closed on its own io exec (TSAN-confirmed data race on MoQSession).

Acquire it through maybeWrapPublisher so the handle is a CrossExecSubscribeNamespaceHandle bound to the session's exec, and give every cross-exec wrapper in PublisherCrossExecFilter a destructor that hops the inner handle's destruction to that exec — so destructor-driven teardown is exec-correct, not just the explicit unsubscribe/cancel calls.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/449)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/449)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #450, opened 2026-06-26
relay: deepen datagram send queue and drop oldest-first

afrind:
Raise the mvfst datagram send queue above the default of 75 and switch the full-queue policy from rejecting the incoming datagram to popping the oldest, so a momentary write stall sheds stale frames rather than failing new writes. Relay listeners use writeBufSize=500; the upstream client leg uses 1000.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/450)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/450)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

gmarzot:
that sounds like an improvement for upstream too

afrind:
> that sounds like an improvement for upstream too

You mean for the moxygen relay?  perhaps.

PR #451, opened 2026-06-26
relay: answer trackStatus without touching registry forwarder in LF mode

afrind:
In LF mode the registry's forwarder lives on the publisher exec, so reading its largest/groupOrder/numForwardingSubscribers from relayExec_ (or the requesting session exec) in trackStatusImpl was an unsynchronized cross-thread read of non-atomic fields.

Resolve trackStatus against thread-local forwarders instead. LocalSubscribeFilter gains a trackStatus override that answers from the requester's local forwarder on the subscriber exec (zero hops). When the requester isn't subscribed, trackStatusImpl hops to the publisher exec and reads the publisher forwarder via tlForwarders_ by FTN, only falling through to upstream when the relay has no forwarding subscription. Non-LF modes still read the registry forwarder inline; relayExec_ owns the sole forwarder there.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/451)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -135,6 +135,14 @@ public:
     );
   }
 
+  folly::coro::Task<TrackStatusResult> trackStatus(moxygen::TrackStatus req) override {
+    // Answer from the local forwarder on this exec; else hop to relayExec_ + upstream.
+    if (auto local = relay_->trackStatusOnSubscriberExec(req)) {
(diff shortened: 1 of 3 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/451)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #452, opened 2026-06-26
relay: fix LF onTrackEvicted no-op and local-forwarder seed-read race

afrind:
In LocalForwarder mode onTrackEvicted looked up a session-keyed subscriber on the registry's publisher forwarder, but the subscriber lives on the per-thread local forwarder (exec-keyed channel subscribers on the publisher fwd) — so it silently evicted nothing and never sent publishDone. Branch on mode(): hop to session's exec, resolve the local forwarder, and evict there.

Also seed a new local forwarder's largest/extensions from a publisherExec hop instead of reading them on subscriberExec while the publisher advances largest_.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/452)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -1110,9 +1110,22 @@ folly::coro::Task<void> MoqxRelay::addSubscriberAndPublishViaLocalForwarder(
     }
   }
 
+  // Capture largest/extensions on publisherExec; reading them on subscriberExec would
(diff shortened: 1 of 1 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/452)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #453, opened 2026-06-26
admin: add GET /config endpoint that dumps the resolved config

afrind:
Adds a /config admin route that serializes the active, resolved Config as JSON (listeners with merged quic/mvfst tunables, services, admin, top-level relay settings). HMAC signing secrets are redacted at the source.

Field enumeration lives in a new format-agnostic walker (config/ConfigSerializer.h) behind a ConfigSink callback interface, so the config library keeps no JSON dependency and the field list has a single home; ConfigHandler.cpp is a thin JsonWriter-backed sink over that walk.

To keep the walker from silently drifting from Config.h, ConfigSerializerTest pins each struct's field count with rfl::num_fields static_asserts (rfl stays out of moqx_config; the check lives in the test target). Adding a field breaks the build until the serializer is updated. Source note in Config.h and a step in docs/dev/config.md point at the serializer.

Tests: ConfigSerializerTest (walk coverage + secret redaction) and test_admin_config.sh (endpoint integration).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/453)


gmarzot (approved):
> @gmarzot reviewed 12 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/453)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna).

Issue #459, opened 2026-06-29
resolveConfig SIGABRTs on empty/unresolvable listener or admin address (should be a clean config error)


gmarzot:
### Summary
`resolveConfig` passes the configured listener/admin `address` straight to `folly::SocketAddress(host, port)`, which throws an uncaught `std::system_error` when the host is empty or unresolvable. The process aborts with **SIGABRT** instead of returning a clean config-validation error.

Same class of bug as #173 (unreadable TLS cert → SIGABRT), but for the `address` field.

### Repro
Minimal config with an empty listener address:
```yaml
listeners:
  - name: main
    udp: {socket: {address: "", port: 9668}}
    tls: {insecure: true}
    endpoint: "/moq-relay"
services:
  s1:
    match: [{authority: {any: true}, path: {exact: "/moq-relay"}}]
    cache: {enabled: true, max_tracks: 100, max_groups_per_track: 3}
```
`moqx validate-config --config empty_addr.yaml`:
```
terminate called after throwing an instance of 'std::system_error'
  what():  Failed to resolve address for '': Name or service not known (error=-2)
*** Signal 6 (SIGABRT) ...
    folly::SocketAddress::getAddrInfo
    openmoq::moqx::config::resolveConfig   (ConfigResolver.cpp)
    openmoq::moqx::config::handleConfigSubcommand  (ConfigInit.cpp:49)
```
An unresolvable hostname (e.g. `address: "no.such.host.invalid"`) hits the same path.

### Cause
`ConfigResolver.cpp` constructs `folly::SocketAddress(sock.address, port)` for listeners (in `resolveListener`) and `folly::SocketAddress(admin.address, port)` for admin, with no check that the address is non-empty/resolvable and no exception handling. folly resolves the host eagerly and throws on failure.

### How it surfaced
Running `docker/config.docker.yaml` (an `entrypoint.sh` template using `${MOQX_BIND_ADDR}`) through `scripts/moqx-run.sh`, which doesn't set `MOQX_BIND_ADDR`; `envsubst` left `address: ""`. The intended path, `docker/entrypoint.sh`, defaults `MOQX_BIND_ADDR=0.0.0.0`, so this only bites when the template is driven by a launcher that doesn't supply that var.

### Suggested fix
[...]

PR #456, opened 2026-06-27
scripts & onboarding cleanup: README + no-sudo, bind-address knob, #441/#391

gmarzot:
Scripts, launcher, and onboarding cleanup. (The empty-address config crash fix moved to its own PR, #461, so this branch stays scripts/docs-only and doesn't overlap the resolver.)

### 1. `scripts/README.md` (new)
Short quickstart for running/benchmarking a moqx relay via `moqx-run.sh` + `config.bench.yaml`: the one-liner, the commands you'll actually use, defaults table, gotchas, and the two-layer logging note (`-x` for moqx's own components, `-v` for the QUIC/HTTP stack).

### 2. Default the relay to no-sudo (`--sudo` to opt in)
Running as yourself is the common case (insecure dev, perf bench, `validate-config`, containers, non-root keys). Sudo is now opt-in via `--sudo`. `perf-test.sh` drops the now-redundant `--no-sudo`.

### 3. Bind-address knob + logging help cleanup
`moqx-run.sh` gains `--bind ADDR` / `MOQX_BIND_ADDR` (default `127.0.0.1`, local-safe). `config.bench.yaml` templates the listener + admin address from it instead of hardcoding `"::"`; `perf-test.sh` passes `--bind ::` so benches keep binding all interfaces (cross-box). The `--help` Logging section now leads with `-x/--xlog` (folly XLOG, primary) and scopes `GLOG_*` as the underlying stack (mvfst/proxygen/fizz); the inert `GLOG_vmodule` default (it named XLOG components, which emit no glog VLOG) is dropped, and a `TODO(folly-xlog)` signpost added.

### 4. Onboarding fixes (#441, #391)
- README/BUILD: `git submodule update --init --recursive` (nested submodules like catapult are required).
- `.gitignore`: ignore `/.docker-deps/` and `AGENTS.md`.

(Note: #441 also mentioned `libbrotli-dev` — intentionally not added. It's optional, not required: picotls auto-detects it via pkg-config and only enables brotli cert compression when present; the build succeeds without it. See the #441 comment.)

---
Related: #461 (empty/unresolvable bind-address crash fix, #459). PKCS#12 PR #460 is stacked on this branch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/456)

PR #460, opened 2026-06-29
tls: PKCS#12 bundle support (in-memory, cloud-vault-friendly)

gmarzot:
Adds support for a single password-protected **PKCS#12 bundle** (`.p12`/`.pfx`) as an alternative to the `cert_file` + `key_file` PEM pair, for both QUIC listeners (mvfst) and the admin HTTPS server. The bundle is decrypted and transcoded to PEM **entirely in memory** — the private key is never written to disk.

### Commits
1. **PKCS#12 in-memory cert/key source** — new `Pkcs12.{h,cpp}` transcode helper (OpenSSL, RAII, password wiped); `pkcs12_file` + `pkcs12_password` / `pkcs12_password_file` on `listener.tls` and `admin.tls`, mutually exclusive with `cert_file`/`key_file`. Resolver transcodes during validation (bad password/file → clean load-time error), stores in-memory `TlsMaterial` on `TlsConfig`. fizz path builds the context from buffers (mirrors the proxygen sample helper so behavior matches the PEM path); admin uses `setCertificateBuf`. picoquic is rejected for now (no in-memory key path wired).
2. **`pkcs12_password_env`** — source the password from a named env var, so a vault/orchestrator can inject it and keep the secret out of the config entirely. `scripts/moqx-run.sh` forwards `MOQX_PKCS12_PASSWORD` to the relay process and across the sudo boundary, never templating it to disk (dry-run shows set/unset, never the value).

### Testing
- `Pkcs12Test` (7): transcode happy path + password-less + wrong-password + empty/garbage + file load/missing.
- `ConfigResolverTest` (+11): mutual exclusion, pico rejection, admin requires cert-or-pkcs12, password-source exclusivity, insecure-ignored + inline-password warnings, env resolution (set/unset), and end-to-end material population for listener + admin.
- All 126 tests green; example config validates; clang-format v19 clean.

### Notes
- **Stacked on #456** (`docs/scripts-readme`) — it shares edits to `MoqxRelayServer.cpp` and `moqx-run.sh`; base will retarget to `main` once #456 merges.
- Security posture and the inline-password caveat are documented in `openmoq/notes/moqx-pkcs12-tls-design.md` (Future direction covers pluggable secret sources and a keyless/async-signer track).
- Unrelated crash surfaced while testing, filed separately as #459 (empty/bad `address` → SIGABRT instead of clean error).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/460)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,6 +228,7 @@ add_library(moqx_config_loader STATIC
   src/config/Loader.cpp
   src/config/ConfigResolver.cpp
   src/config/ConfigInit.cpp
+  src/config/Pkcs12.cpp
 )
 
 target_include_directories(moqx_config_loader
@@ -241,6 +242,9 @@ target_link_libraries(moqx_config_loader PUBLIC
   yaml-cpp
   Folly::folly_file_util
 )
+
+# PKCS#12 transcode (Pkcs12.cpp) + OPENSSL_cleanse in ConfigResolver.cpp.
+target_link_libraries(moqx_config_loader PRIVATE OpenSSL::Crypto)
 
 target_compile_options(moqx_config_loader PRIVATE -Wall -Wextra -Wpedantic)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,8 +243,12 @@ target_link_libraries(moqx_config_loader PUBLIC
   Folly::folly_file_util
 )
 
-# PKCS#12 transcode (Pkcs12.cpp) + OPENSSL_cleanse in ConfigResolver.cpp.
-target_link_libraries(moqx_config_loader PRIVATE OpenSSL::Crypto)
+# PKCS#12 transcode (Pkcs12.cpp) + OPENSSL_cleanse in ConfigResolver.cpp;
+# folly cert utils supplies OpenSSLCertUtils::pemEncode.
+target_link_libraries(moqx_config_loader PRIVATE
+  OpenSSL::Crypto
+  Folly::folly_ssl_openssl_cert_utils
+)
 
 target_compile_options(moqx_config_loader PRIVATE -Wall -Wextra -Wpedantic)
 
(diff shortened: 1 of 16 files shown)


afrind (approved):
> @afrind reviewed 16 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/460)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna, peterchave, suhasHere, TimEvens, and wilaw).

___
*[`src/MoqxRelayServer.cpp` line 43 at r1](https://reviewable.io/reviews/openmoq/moqx/460#-OwUgELhESa6v_NgcFML:-OwUgELhESa6v_NgcFMM:b-jfyhba) ([raw file](https://github.com/openmoq/moqx/blob/4256841b3e9e3e30fdd0bf89ae3ea67c1b3bd50e/src/MoqxRelayServer.cpp#L43)):*
> ```cpp
> // Build a FizzServerContext from in-memory PEM buffers (cert chain + key)
> // instead of file paths, so a PKCS#12-derived private key never touches disk.
> // This mirrors quic::samples::createFizzServerContextImpl (proxygen hq sample,
> ```

Do we need this detail here?
___
*[`src/MoqxRelayServer.cpp` line 46 at r2](https://reviewable.io/reviews/openmoq/moqx/460#-OwUrrOb4nWEEcaA_7aJ:-OwUrrOcB7O7YH7IAoZs:bqz1ph5) ([raw file](https://github.com/openmoq/moqx/blob/746a0fd57387a298d7193580e323026684e845a7/src/MoqxRelayServer.cpp#L46)):*
> ```cpp
> // minus the insecure-default branch) so a pkcs12-sourced context behaves
> // identically to the PEM-file one (session tickets, resumption, ALPN mode).
> // Keep in sync with that helper if it changes upstream. TODO: replace with a
> ```

File a follow up task

Issue #441, opened 2026-06-24
Update Quick Start instructions


peterchave:
Minor tweaks to the Quick Start instructions.

Testing against a vanilla Ubuntu 22.04 LTS machine:

Submodule update should be recursive:
```git submodule update --init --recursive```

libbrotli-dev is missing on a vanilla Ubuntu machine add note to install or should this be added to the system-deps script?
```sudo apt install libbrotli-dev```

gmarzot:
On the two items here:

**Recursive submodule** — done in #456 (`git submodule update --init --recursive` in README + BUILD).

**`libbrotli-dev`** — after tracing it, this is **optional, not a build requirement**, so we're intentionally *not* adding it as a required dep (see #466, which removes a misleading hint):

- The only consumer is **picotls** (pulled in by the pico transport), which probes brotli via `PKG_CHECK_MODULES(BROTLI_DEC libbrotlidec)` and enables brotli cert compression **only if found**. Without it, the build succeeds — just without brotli cert compression. folly here is built with brotli disabled; fizz/proxygen/mvfst don't use it.
- It's therefore **not an upstream moxygen / `install-system-deps.sh` concern** either (upstream has no picotls).

Likely why some hit it and others didn't: it's a **partial-state × from-source interaction**. The `.pc` files ship only with `libbrotli-dev`, but `libbrotli1` (runtime `.so`) is often already present transitively on desktop Ubuntu. If a `--from-source` configure **half-detects** brotli (enough to turn it on) but the dev headers/linker symlink aren't there, the build fails mid-way and installing `libbrotli-dev` "fixes" it — whereas a truly clean machine skips brotli cleanly and builds fine.

If from-source builders keep tripping on this, the robust fix is to make brotli **deterministic** (explicitly disable, or document it as an optional dep so detection always succeeds) rather than imply it's required. Happy to do that as a follow-up if desired.

gmarzot:
Resolved by #456 (merged): `git submodule update --init --recursive` in README + BUILD.

The `libbrotli-dev` part is intentionally **not** added — it's optional, not required: picotls auto-detects it via pkg-config and only enables brotli cert compression when present; the build succeeds without it (and it's picotls-specific, not a moxygen/upstream concern). #466 removes the misleading hint, and the detailed finding (incl. why some from-source builds tripped on it) is in the comment above. Closing as addressed; happy to do a deterministic-brotli follow-up if from-source builders keep hitting it.

PR #466, opened 2026-06-30
build.sh: drop libbrotli-dev from apt hint (optional, not required)

gmarzot:
Follow-up to #456. Removes `libbrotli-dev` from the Ubuntu/Debian apt hint in `scripts/build.sh`.

It was added implying it's a required system dep, but it isn't:
- **Optional, auto-detected.** picotls (pulled in by the pico transport) probes it via `PKG_CHECK_MODULES(BROTLI_DEC libbrotlidec)` and only enables brotli cert compression (`-DPICOTLS_USE_BROTLI=1`) when found. Without it, the build succeeds — just without brotli cert compression.
- **Not a moxygen/upstream concern.** folly here is built with brotli disabled (no `FOLLY_HAVE_BROTLI`); fizz/proxygen/mvfst don't reference it. The dependency is picotls-specific (openmoq's pico transport), so it's not relevant to upstream facebookexperimental/moxygen's `install-system-deps.sh`.

Refs #441 (the original request — its `libbrotli-dev` ask is addressed here as "optional, not added"; the recursive-submodule part already landed in #456).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/466)

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -180,7 +180,7 @@ check_system_deps() {
           echo "  sudo apt-get install -y build-essential cmake ninja-build \\"
           echo "    libssl-dev libunwind-dev libgoogle-glog-dev libgflags-dev \\"
           echo "    libdouble-conversion-dev libevent-dev libsodium-dev libzstd-dev \\"
-          echo "    libboost-all-dev libfmt-dev zlib1g-dev libc-ares-dev gperf libbrotli-dev"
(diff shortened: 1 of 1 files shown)


afrind (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/466)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on peterchave and suhasHere).

PR #473, opened 2026-07-01
perf-test.sh: LD_PRELOAD jemalloc for the perf client too

afrind:
Previously -j/--jemalloc only preloaded jemalloc for the relay (via moqx-run). The perf client is launched directly, so resolve libjemalloc.so.2 ourselves and LD_PRELOAD it for the client as well, mirroring moqx-run's candidate list + ldconfig fallback. For --remote-client, resolve the lib on the remote host since the path is host-specific.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/473)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/473)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #478, opened 2026-07-01
stats: widen object-ack latency histogram tail to 1s

afrind:
Give object-ack latency its own bucket array starting at 500us and extending through 250ms/500ms/1s, instead of sharing kLatencyBucketsUs which tops out at 100ms. Acks stretch well past 100ms under loss, so the old top bucket swallowed the tail.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/478)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/478)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #485, opened 2026-07-02
build: support Ubuntu/Debian derivatives natively, fix from-source setup

mondain:
Makes the native (non-Docker) build path work beyond exactly Ubuntu 22.04 / Debian 12. Companion moxygen PR: openmoq/moxygen#309 (ID_LIKE installer dispatch, ubuntu-24.04-amd64 tarballs, CONFIG-mode gtest probe).

## Problem

Native builds only worked on Ubuntu 22.04, Debian 12, or macOS 15:

- Platform detection switched on os-release `ID` alone, so derivatives (Linux Mint reports `ID=linuxmint`) died at step one with "unsupported Linux distro" — in both the tarball fetcher and the dev-artifact fallback (which had a second, divergent detector).
- `check_system_deps` passed even when required boost component libs were absent, leaving an opaque `boost_context-config.cmake not found` configure failure instead of naming the missing packages.
- The from-source fallback configured moxygen while a stale tarball install was still present at `CMAKE_INSTALL_PREFIX` (which CMake searches implicitly), so the gtest probe found the tarball's GTestConfig.cmake, skipped bundling GoogleTest, deleted that tree at install time, and left moqx unable to configure (`find_package(GTest REQUIRED CONFIG)` failure).

## Changes

- New `scripts/detect-platform.sh`, shared by `setup-deps-tarball.sh` and `setup-deps-dev-artifact.sh`. Honors `ID_LIKE`: Ubuntu derivatives map to the Ubuntu base release via `UBUNTU_CODENAME` (or `/etc/upstream-release/lsb-release`), Debian derivatives map to bookworm. Errors suggest the `MOQX_PLATFORM` override.
- `build.sh check_system_deps`: probe each required boost component (context, filesystem, program_options, regex, thread) and the double-conversion header; missing deps now surface as named apt packages before configure.
- `setup-deps-standalone.sh`: wipe the install tree before configuring, not just before installing.
- BUILD.md: document derivative support.

## Verification

On Linux Mint 22.3 (noble base): detection resolves `ubuntu-24.04-amd64`; tarball fetch fails over cleanly (the asset exists once openmoq/moxygen#309 lands and the snapshot regenerates); from-source setup, full build, and test suite run. 691/711 tests pass; the 20 failures are UpstreamProviderTest/RelayUpstreamSubscribeRaceTest QUIC-loopback connect timeouts that occur on this host in both dependency modes — environmental, unrelated to this diff (no C++ touched).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/485)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -11,6 +11,7 @@ formatting, and CI.
 |----------|--------|-------|
 | Ubuntu 22.04 (Jammy) | Tested in CI | Primary dev/CI platform |
 | Debian 12 (Bookworm) | Tested (Docker build) | Docker image base |
+| Ubuntu/Debian derivatives (Mint, Pop!_OS, ...) | Supported | Detected via `ID_LIKE`, mapped to the Ubuntu/Debian base artifacts; override with `MOQX_PLATFORM` if detection is wrong |
 | Fedora / RHEL | TBD | `install-system-deps.sh` has dnf support; not yet CI-tested |
 | macOS (Homebrew) | TBD | `install-system-deps.sh` has brew support; not yet CI-tested |
 
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -151,6 +151,28 @@ check_system_deps() {
     fi
   fi
 
+  # Boost component libs — packaged separately on Debian/Ubuntu (headers via
+  # libboost-dev do NOT include them). folly's cmake config find_package()s
+  # each component, so a missing one fails configure with an opaque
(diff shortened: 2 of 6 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/485)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind and peterchave).

PR #497, opened 2026-07-20
moq_decode: parse draft-18 SETUP and name extension types

afrind:
Add a SETUP (0x2f00) parser and unify it with extension parsing in a shared count-less delta-KV walker (p_options) that takes the label registry as an argument. Setup options and object/Track-Property extensions share the same wire grammar; unknown keys are extension points in both. Name the extension/property types (DELIVERY_TIMEOUT, etc.) and flatten the immutable-extensions container (0xB) with a visible block boundary. p_params stays count-based for Request/Track params (strict unknowns, value shapes).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/497)

### DIFF
--- a/tools/moq_decode.py
+++ b/tools/moq_decode.py
@@ -67,6 +67,25 @@ SETUP_PARAM_KEYS = {
     0: "ROLE",
     1: "PATH",
     2: "MAX_REQUEST_ID",
+    3: "AUTHORIZATION_TOKEN",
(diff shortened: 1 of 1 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/497)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #520, opened 2026-07-27
tools: add per-scrape/averaging window control for histogram metrics

afrind:
Histogram-derived metrics (evb loop busy, object ACK latency, client latency) now compute p99/p50/mean over the per-scrape counter delta by default, with a new "avg" dropdown to average over a 10s/30s/1m/5m span instead. histWindowStats() diffs against the previous scrape when the span is 0, else against the oldest snapshot within the span, and returns an empty stat on the first scrape when no delta exists yet.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/520)

### DIFF
--- a/tools/metrics-dashboard.html
+++ b/tools/metrics-dashboard.html
@@ -111,6 +111,15 @@
       <option value="3600">1h</option>
     </select>
   </label>
+  <label class="status">avg
+    <select id="avgWindow">
+      <option value="0" selected>per-scrape</option>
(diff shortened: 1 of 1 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/520)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #521, opened 2026-07-27
relay: give subscribe-initiated forwarders the LF publisher chain + tl slot

afrind:
In LF mode a subscribe-triggered upstream subscription created its publisher forwarder on relayExec_ with a MoqxRelay-direct callback and never placed it in tlForwarders_. That forwarder lives on the upstream session's iothread, so its onEmpty/forwardChanged/newGroupRequested ran there and touched registry_ off relayExec_, and same-thread subscribers missed the zero-hop fast path.

Factor the publisher-forwarder callback chain (Weak -> CrossExec -> Local) and tlForwarders_ slot claim out of createPublisherForwarder into a shared installPublisherForwarderCallbackChain helper. The first subscriber now installs that same Weak->CrossExec(relayExec_)->Local chain and claims the tl slot on the publisher's exec (removeOnEmpty=true, since subscribe-initiated tracks drop on empty). getOrCreateFromSubscribe accepts a null callback so the LF caller installs the real chain on the forwarder's exec; non-LF still passes the relay directly.

Document the resulting symmetry in the callback-chain overview and dev doc: subscribe-initiated now uses the same chain as publish-initiated, with the removeOnEmpty contract differing (true vs false).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/521)

### DIFF
--- a/docs/dev/local-forwarder-flow.md
+++ b/docs/dev/local-forwarder-flow.md
@@ -135,6 +135,8 @@ upstream work, and nests a single sortie to `[Pub]` to wire the channel sub.
 [Relay]                 ├─▶ joinOrPrepareUpstreamSubscription()  # registry: first vs subsequent
 [Relay]                 ├─▶ buildLocalToPublisherCallbacks()
 [Relay]                 └─▶ ⇢⇢▶ [Pub]  single sortie:
+[Pub]                        ├─ if FIRST subscriber:
+[Pub]                        │    └─▶ installPublisherForwarderCallbackChain()  # chain + tlForwarders_ slot
 [Pub]                        ├─▶ installChannelSubscriber(localFwd ↔ publisherFwd)
 [Pub]                        └─ if FIRST subscriber:
 [Pub]                             ├─▶ addChannelSubscriber(relayChain, passive)
@@ -151,9 +153,20 @@ upstream work, and nests a single sortie to `[Pub]` to wire the channel sub.
 ```
 
 The first subscriber does the heavy lifting (upstream subscribe + installing the relay chain,
-the same passive cache/Top-N chain described in [Data flow](#data-flow)). Subsequent
-subscribers on the same thread hit the `attachSubscriber` fast path; on other threads they take
-only the `installChannelSubscriber` half of the sortie.
+the same passive cache/Top-N chain described in [Data flow](#data-flow)). It also installs the
+publisher-forwarder control chain and claims the `tlForwarders_` slot via
+`installPublisherForwarderCallbackChain` — the **same** wiring the publish path uses, so a
+subscribe-initiated publisher forwarder is symmetric with a publish-initiated one. Subsequent
+subscribers on the same thread hit the `attachSubscriber` fast path (now also for
+subscribe-initiated tracks); on other threads they take only the `installChannelSubscriber` half
+of the sortie.
+
+The forwarder is created on `relayExec_` (in `joinOrPrepareUpstreamSubscription`) with a **null**
+callback, then gets its real callback + tl slot installed on `[Pub]` in the first-subscriber
+sortie — `tlForwarders_.get()` must run on the forwarder's own exec. It uses `removeOnEmpty=true`
+(subscribe-initiated tracks drop on empty, unlike publish): `LocalForwarderCallback` vacates the
+tl slot on `[Pub]` when the last subscriber leaves, while `onEmptyImpl` unsubscribes upstream and
+removes the registry entry on `[Relay]`.
 
 ### Why each guard exists
 
@@ -250,8 +263,10 @@ The reusable adapter layers:
 
 ### Publisher forwarder → relay state
 
-Built by `createPublisherForwarder`. Lifecycle events on the publisher's own forwarder must
-reach relay-global state on `relayExec_`:
+Built by `installPublisherForwarderCallbackChain` for both publish-initiated (via
+`createPublisherForwarder`) and subscribe-initiated (via `attachNewLocalForwarderOnRelayExec`'s
+first-subscriber sortie) tracks. Lifecycle events on the publisher's own forwarder must reach
+relay-global state on `relayExec_`:
 
 ```
 [Pub]  publisherFwd fires onEmpty / forwardChanged / newGroupRequested
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -14,6 +14,7 @@
 #include "relay/PublisherCrossExecFilter.h"
 #include "relay/SubscriberCrossExecFilter.h"
 #include "relay/WeakRelayForwarderCallback.h"
+#include <folly/ScopeGuard.h>
 #include <folly/container/F14Set.h>
 #include <moxygen/MoQFilters.h>
 #include <moxygen/MoQTrackProperties.h>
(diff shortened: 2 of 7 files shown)


afrind (commented):
> @afrind made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/521)* status: 0 of 8 files reviewed, 5 unresolved discussions.

___
*[`src/MoqxRelay.cpp` line 479 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_W11O4N6PdqSyy6Qp:-Oy_W11P1I5CSQmkFtY5:bavgrst) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/MoqxRelay.cpp#L479)):*
> ```cpp
>   // Authoritative slot claim; displaces any stale subscribe-path local forwarder so
>   // same-thread subscribers reuse THIS forwarder via the fast path. Publish forwarders
>   // (removeOnEmpty=false) are seeded at creation; subscribe claims await upstream OK.
> ```

Seeded is used without definition here, so it's a bit confusing.  I'm loath to make this comment longer, but as is it's not readable.
___
*[`src/SubscriptionRegistry.cpp` line 51 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_WWxQ7ZRhlTa-Utc-:-Oy_WWxQ7ZRhlTa-Utc0:b-li2zzm) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/SubscriptionRegistry.cpp#L51)):*
> ```cpp
>     auto forwarder = std::make_shared<moxygen::MoQForwarder>(ftn, largest);
>     // Null callback: caller installs the real chain later on the forwarder's exec (LF path).
>     if (callback) {
> ```

Is setting a null callback on forwarder harmful?  It might be better to set it unilaterally and make no changes to SubscriptionRegistry.
___
*[`src/MoqxRelay.cpp` line 1983 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_WLKI-cFWzLl4piw3:-Oy_WLKI-cFWzLl4piw4:b-gie0k4) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/MoqxRelay.cpp#L1983)):*
> ```cpp
>   }
> 
>   // Fulfill on every exit (incl. error/cancel below) so isNew=false waiters never hang.
> ```

Do waiters need to differentiate between a successful exit and an unsuccessful one?  We _could_ used readiness->setException?
___
*[`src/relay/LocalForwarderCallback.h` line 25 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_WlJR9lbyYqo93h-c:-Oy_WlJR9lbyYqo93h-d:b6lg3f2) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/relay/LocalForwarderCallback.h#L25)):*
> ```objc
> //
> // removeOnEmpty distinguishes the two roles:
> //   - subscribe-path local & subscribe-initiated publisher forwarders (removeOnEmpty=true): when
> ```

Is this change adding value?
___
*[`src/relay/LocalForwarderRegistry.h` line 33 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_Wtsv9MTSA3Zow4Xp:-Oy_Wtsv9MTSA3Zow4Xq:bt7ngg9) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/relay/LocalForwarderRegistry.h#L33)):*
> ```objc
>   struct Entry {
>     std::shared_ptr<moxygen::MoQForwarder> forwarder;
>     std::shared_ptr<folly::SharedPromise<folly::Unit>> readiness;
> ```

Would this read better as "ready" rather than "readiness"?
___
*[`src/relay/LocalForwarderRegistry.h` line 56 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_X5sQ5NL6I9yGKYXx:-Oy_X5sQ5NL6I9yGKYXy:b-1usx48) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/relay/LocalForwarderRegistry.h#L56)):*
> ```objc
>   // here, and drains itself via the source-termination cascade (its identity-
>   // checked removal then no-ops, since this forwarder now owns the slot).
>   // seeded=true releases readiness waiters now (the claiming forwarder's largest is
> ```

maybe rename seeded to largestKnown or something similar?

michalhosna (commented):
I (and claude) see races. 

See https://github.com/openmoq/moqx/commit/43425c1fb2c52765d012d6c1a8d8d45b0de07208

> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/521#-:-OydVplS39sXCuEueJvX:b-uen5c7)* status: 0 of 8 files reviewed, 5 unresolved discussions.

afrind (commented):
> @afrind made 4 comments and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/521)* status: 0 of 8 files reviewed, 1 unresolved discussion.

___
*[`src/MoqxRelay.cpp` line 1983 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_WLKI-cFWzLl4piw3:-OyikSJR6nIyIgtDgzgn:b-886gsx) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/MoqxRelay.cpp#L1983)):*

Do waiters need to differentiate between a successful exit and an unsuccessful one?  We _could_ used readiness->setException?
</blockquote></details>

Fixed
___
*[`src/relay/LocalForwarderCallback.h` line 25 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_WlJR9lbyYqo93h-c:-OyikUky8_J_MHW6frSE:b-886gsx) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/relay/LocalForwarderCallback.h#L25)):*

Is this change adding value?
</blockquote></details>

Fixed
___
*[`src/relay/LocalForwarderRegistry.h` line 56 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_X5sQ5NL6I9yGKYXx:-OyikWqF5FiIirpr5rzG:blzx09u) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/relay/LocalForwarderRegistry.h#L56)):*

maybe rename seeded to largestKnown or something similar?
</blockquote></details>

Decided on ready (since largest is unknown when no content published yet)
___
*[`src/SubscriptionRegistry.cpp` line 51 at r1](https://reviewable.io/reviews/openmoq/moqx/521#-Oy_WWxQ7ZRhlTa-Utc-:-Oyikc9e37PPxzMlEZwS:b-886gsx) ([raw file](https://github.com/openmoq/moqx/blob/9bc3b551817f0e693b55ca5601281e5d1e6ce6cd/src/SubscriptionRegistry.cpp#L51)):*

Is setting a null callback on forwarder harmful?  It might be better to set it unilaterally and make no changes to SubscriptionRegistry.
</blockquote></details>

Fixed

afrind:
replaced this with the stack at #541

PR #522, opened 2026-07-27
relay: move LF forwarder ownership to the publisher exec, off the registry

afrind:
In LocalForwarder mode the publisher forwarder is owned by tlForwarders_ on the publisher exec; clear the registry's ref once that ownership transfers. Callers resolve the forwarder by FullTrackName from the publisher exec rather than holding a registry ref, and duplicate-publisher eviction drains the old forwarder on its own exec (via the displaced same-exec ref or tlForwarders_).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/522)

### DIFF
--- a/docs/dev/local-forwarder-flow.md
+++ b/docs/dev/local-forwarder-flow.md
@@ -48,9 +48,13 @@ Executor legend used below:
 
 ## Publish
 
-A publisher's local forwarder **is** the publisher forwarder — the same object serves as the
-on-thread data-plane forwarder and as the registry's publisher entry (no second forwarder is
-constructed on `relayExec_`, unlike the `RelayExec`-mode branch in `publishWithSession`).
+A publisher's local forwarder **is** the publisher forwarder — a single forwarder serves the
+whole track, living on the publisher's executor rather than on `relayExec_`. Ownership lives in
+`tlForwarders_` on the publisher's executor, **not** in the registry: once registration
+completes, `registerPublishOnRelayExec` calls `registry_.clearForwarder(ftn)`, so the registry
+entry keeps its subscription/namespace bookkeeping but drops its forwarder ref. Code that needs
+the forwarder resolves it by `FullTrackName` from `tlForwarders_` on the owning executor, rather
+than holding a registry ref.
 
 ### Setup
 
@@ -65,7 +69,8 @@ constructed on `relayExec_`, unlike the `RelayExec`-mode branch in `publishWithS
 [Relay]            └─▶ registerPublishOnRelayExec()
 [Relay]                 ├─▶ publishWithSession()              # registry + namespace tree
 [Relay]                 │     └─▶ addSubscriberAndPublish()   # attach existing SUBSCRIBE_NAMESPACE subs (see Publish fan-out)
-[Relay]                 └─▶ relayChainFilter->setDownstream(topNFilter)
+[Relay]                 ├─▶ relayChainFilter->setDownstream(topNFilter)
+[Relay]                 └─▶ registry_.clearForwarder(ftn)     # drop registry ref; tlForwarders_ owns the fwd
 [Pub]         return {consumer, replyTask}                    # immediate
 ```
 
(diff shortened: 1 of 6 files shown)


afrind (commented):
> @afrind made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/522)* status: 0 of 6 files reviewed, 4 unresolved discussions.

___
*[`src/MoqxRelay.h` line 490 at r1](https://reviewable.io/reviews/openmoq/moqx/522#-Oy_loUZ6TMzZGgQG3dg:-Oy_loUZ6TMzZGgQG3dh:b5p4yp0) ([raw file](https://github.com/openmoq/moqx/blob/a537cd2ce312024d0ec9782561abce9d74b841fb/src/MoqxRelay.h#L490)):*
> ```objc
>     // Subsequent-subscriber seed (set only when !ownsRelayChain): the live publisher
>     // forwarder's largest/extensions, read on the publisher exec.
>     std::optional<moxygen::AbsoluteLocation> seedLargest;
> ```

I don't like "seed" here
___
*[`src/MoqxRelay.cpp` line 651 at r1](https://reviewable.io/reviews/openmoq/moqx/522#-Oy_mcYUCpXyRjoCKwvI:-Oy_mcYUCpXyRjoCKwvJ:b-z977ug) ([raw file](https://github.com/openmoq/moqx/blob/a537cd2ce312024d0ec9782561abce9d74b841fb/src/MoqxRelay.cpp#L651)):*
> ```cpp
>     // Null handle => previous publisher already terminated and onPublishDone() tore it down; skip.
>     if (evicted.handle && mode() == Mode::LocalForwarder) {
>       // The old forwarder lives on its own exec (registry no longer holds it); unsubscribe and
> ```

We don't need comments explaining what we no longer do.  Also "its own exec" "its publisher's exec"
___
*[`src/MoqxRelay.cpp` line 660 at r1](https://reviewable.io/reviews/openmoq/moqx/522#-Oy_n4jmCA99Y_ULZyGA:-Oy_n4jmCA99Y_ULZyGB:b-pjqxsc) ([raw file](https://github.com/openmoq/moqx/blob/a537cd2ce312024d0ec9782561abce9d74b841fb/src/MoqxRelay.cpp#L660)):*
> ```cpp
>             auto oldFwd = displacedLocalFwd;
>             if (!oldFwd) {
>               auto* localReg = tlForwarders_.get();
> ```

Add a one line comment explaining this case in a bit more detail.  You can remove "using the displaced ref (same-exec) or tlForwarders_ (cross-exec)." above, in favor of a bit more detail here.
___
*[`src/MoqxRelay.cpp` line 885 at r1](https://reviewable.io/reviews/openmoq/moqx/522#-Oy_nJk8A6sNSQIGATpw:-Oy_nJk8A6sNSQIGATpx:bis4yur) ([raw file](https://github.com/openmoq/moqx/blob/a537cd2ce312024d0ec9782561abce9d74b841fb/src/MoqxRelay.cpp#L885)):*
> ```cpp
>   }
>   if (!forwarder) {
>     return false;
> ```

Does this case warrant an error log?
___
*[`src/MoqxRelay.cpp` line 919 at r1](https://reviewable.io/reviews/openmoq/moqx/522#-Oy_nRGR8U-T89RDJXxP:-Oy_nRGR8U-T89RDJXxQ:b52i5l8) ([raw file](https://github.com/openmoq/moqx/blob/a537cd2ce312024d0ec9782561abce9d74b841fb/src/MoqxRelay.cpp#L919)):*
> ```cpp
>         publisherExec,
>         [pf = std::move(publisherFwd), ex = subscriberExec, re = relayExec]() noexcept {
>           // lock on the pub exec: resolves the exact wired forwarder or null if it is gone.
> ```

You can add another line to help explain why it can be gone
___
*[`src/MoqxRelay.cpp` line 2163 at r1](https://reviewable.io/reviews/openmoq/moqx/522#-Oy_oY1zBbbnjjIQZqbt:-Oy_oY2-2dry4JMan41W:b-3d57dy) ([raw file](https://github.com/openmoq/moqx/blob/a537cd2ce312024d0ec9782561abce9d74b841fb/src/MoqxRelay.cpp#L2163)):*
> ```cpp
>   } else {
>     // Subsequent subscriber, first on this iothread: seed from the live publisher
>     // forwarder so the OK carries the established largest, not this fresh forwarder's
> ```

Remove ", not this fresh forwarder's
    // unseeded value a client reads as a track restart.
"
___
*[`src/MoqxRelay.cpp` line 2748 at r1](https://reviewable.io/reviews/openmoq/moqx/522#-Oy_oeh6BinJvzXnpqpw:-Oy_oeh6BinJvzXnpqpx:b-shhjbo) ([raw file](https://github.com/openmoq/moqx/blob/a537cd2ce312024d0ec9782561abce9d74b841fb/src/MoqxRelay.cpp#L2748)):*
> ```cpp
>         .sourceAddress = sourceAddr,
>     };
>     // LF mode clears the registry forwarder (owned by tlForwarders_); stats unavailable here.
> ```

FYI - this breaks some components of the /state endpoint that were never really safe in MT mode anyways.  We can file an issue to circle back and fix

afrind:
I'm going to replace this with a different approach

PR #523, opened 2026-07-27
scripts: add -j/--jobs and MOQX_BUILD_JOBS to control build parallelism

afrind:
Both build.sh and setup-deps-standalone.sh hardcoded nproc for the cmake --build job count. distcc wants far more jobs than the local core count, so add an explicit override with precedence: -j flag, then MOQX_BUILD_JOBS, then nproc.

Also loosen the optional-SHA lookahead in --from-release/--from-source from "$1" != --* to "$1" != -* so a following -j isn't consumed as a SHA.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/523)

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -6,8 +6,8 @@
 #   --from-source   — build moxygen + all Meta deps from source (~15-30 min)
 #
 # Usage:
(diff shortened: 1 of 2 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/523)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna).

PR #533, opened 2026-07-31
stats: per-track counters exposed at /metrics/track

afrind:
Per-(track, iothread) TrackStats live in the stats tier, owned by a TrackStatsRegistry that binds one TrackStatsCollector per data-plane thread and merges them on demand — the same split as StatsRegistry/MoQStatsCollector. A TrackStatsFilter counts on the relay ingest chain and on each downstream subscriber, including the PUBLISH fan-out path. No read path dereferences a registry forwarder, so counting behaves identically in single-thread, relay-exec and local-forwarder modes.

GET /metrics/track?service=&namespace=&track=&limit= reports subscribers, groups/subgroups/objects/datagrams/bytes in both directions, and publish-start/last-object timestamps. Rates are left to Prometheus. A match wider than the limit is rejected rather than truncated: an arbitrary F14-ordered subset would reshuffle between scrapes.

Namespaces and track names are safeName()-encoded in labels and expected in the same form on the query, so a scraped value pastes back into a query and two tracks cannot collapse onto one label set.

Groups are counted against an LRU window of the 3 most recent group IDs, so subgroups of concurrently-open groups do not inflate the count.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)


gmarzot (approved):
> @gmarzot reviewed 29 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

gmarzot:
image

claude is happy w/ it it at least and it will populate real data in a dashboard now, but without a way to iterate all (the wrong model anyway) the production deploy utility is limited. I am thinking about a more complete scalable push model ... M4 perhaps

image

afrind (commented):
> @afrind made 23 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)* status: 18 of 29 files reviewed, 2 unresolved discussions (waiting on gmarzot).

___
*[`docs/config.md` line 436 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7hX_rCXwsMyxGvOXO:-Oz7hX_rCXwsMyxGvOXP:bc8npk) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/config.md#L436)):*
> ```Markdown
>   #   alpn: [h2, "http/1.1"]
>   # track_metrics_enabled: true    # per-track counting and /metrics/track
>   # track_metrics_limit: 10        # default tracks per /metrics/track scrape
> ```

track_metrics_endpoint_default_limit
___
*[`docs/config.md` line 437 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7hbAx4xpqTnUmsbBW:-Oz7hbAx4xpqTnUmsbBX:b-lrzxjp) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/config.md#L437)):*
> ```Markdown
>   # track_metrics_enabled: true    # per-track counting and /metrics/track
>   # track_metrics_limit: 10        # default tracks per /metrics/track scrape
>   # track_metrics_max_limit: 1000  # ceiling on the ?limit= parameter
> ```

track_metrics_endpoint_max_limit
___
*[`docs/metrics.md` line 222 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7hvuz6g13qJL_VtTH:-Oz7hvuz6g13qJL_VtTI:b-9iv8ji) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/metrics.md#L222)):*
> ```Markdown
> `conf.2eexample.2ecom-room.201`.
> 
> Because the encoding is injective, two tracks can never collapse onto one label
> ```

Remove the "Because the encoding is injective, " bit
___
*[`docs/metrics.md` line 246 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7i74_-ohZt3RF9dUi:-Oz7i74_-ohZt3RF9dUj:b-k03zki) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/metrics.md#L246)):*
> ```Markdown
> concurrently-open groups can arrive interleaved without inflating the count. A
> group revisited after 3 *other* groups have been seen is counted again — the
> window bounds per-filter state, and MoQ group IDs advance, so this only shows up
> ```

and MOQT group IDs usually advance and high concurrency is not expected,
___
*[`docs/metrics.md` line 249 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7iJyN0Zc_6aNFZMqp:-Oz7iJyN0Zc_6aNFZMqq:b-61okqc) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/metrics.md#L249)):*
> ```Markdown
> in pathological interleaving.
> 
> Byte counters measure **object payload bytes**, not wire bytes: no MOQT headers,
> ```

I'll file a follow up issue to report header bytes.
___
*[`docs/metrics.md` line 254 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7ic9a3LtSZ0xI97kD:-Oz7ic9a3LtSZ0xI97kE:b-kw8kcd) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/metrics.md#L254)):*
> ```Markdown
> 
> Sent counters are summed across subscribers, so one ingested object fanned out to
> three subscribers increments `objects_sent` by 3.
> ```

Should note that "sent" here means "passed to transport" and e.g. and objects that are not sent due to stop_sending, delivery timeout, etc aren't reflected in the counts.
___
*[`docs/metrics.md` line 258 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7irg74-rC0ZEJqbjd:-Oz7irg74-rC0ZEJqbje:b-4d0r09) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/metrics.md#L258)):*
> ```Markdown
> ### Bitrate and object rate
> 
> Rates are not computed server-side — derive them from the counters:
> ```

Do we need to say that we don't support rates?  This section could be removed.
___
*[`src/MoqxRelay.cpp` line 2616 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7jAwk-PJiJoF5jaoL:-Oz7jAwk-PJiJoF5jaoM:b-8z02ab) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/MoqxRelay.cpp#L2616)):*
> ```cpp
>     const {
>   TrackMatch match;
>   // forEachName, not forEach: EntryView copies two shared_ptrs per entry, which
> ```

This comment is not helpful. remove
___
*[`src/MoqxRelayContext.cpp` line 177 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7jMlcBE-sRFzup1yI:-Oz7jMlcBE-sRFzup1yJ:blwxcnd) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/MoqxRelayContext.cpp#L177)):*
> ```cpp
> }
> 
> void MoqxRelayContext::initThreadStats(folly::IOThreadPoolExecutor& ioExecutor, bool trackStats) {
> ```

call this initTrackStats
___
*[`src/MoqxRelayContext.cpp` line 328 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7jrwC1iLT7oihZkuz:-Oz7jrwC1iLT7oihZkv-:b-b0tzwc) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/MoqxRelayContext.cpp#L328)):*
> ```cpp
> 
>   // Each surviving service costs one executor hop per thread below.
>   std::erase_if(services, [](const Service& s) { return s.match.keys.empty(); });
> ```

Do this as an if in the loop below, omitting an entry in countTasks?
___
*[`src/MoqxRelayContext.cpp` line 338 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7k7--E0kzVJe177vN:-Oz7k7--E0kzVJe177vO:b-6avm5f) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/MoqxRelayContext.cpp#L338)):*
> ```cpp
>   auto counters = co_await folly::coro::collectAllRange(std::move(countTasks));
>   for (size_t i = 0; i < services.size(); ++i) {
>     for (const auto& ftn : services[i].match.keys) {
> ```

Why are we scanning keys here instead of just iterating counters and pushing back service it->first, it->second?
___
*[`src/SubscriptionRegistry.h` line 182 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7kdHa1DuLBnQLGXuq:-Oz7kdHa1DuLBnQLGXur:b8qqtel) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/SubscriptionRegistry.h#L182)):*
> ```objc
>   void forEach(folly::FunctionRef<void(const EntryView&)> fn) const;
> 
>   // Cheaper than forEach for callers that only need names: EntryView copies a
> ```

This comment is more or less obvious, remove?
___
*[`src/admin/TrackMetricsHandler.h` line 19 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7kptE12h6OaiWQ4WU:-Oz7kptE12h6OaiWQ4WV:b-9ffd4s) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.h#L19)):*
> ```objc
> class AdminServer;
> 
> struct TrackMetricsLimits {
> ```

this is sort of the same as the conig struct?  Is it a normal pattern to include it here?
___
*[`src/admin/TrackMetricsHandler.cpp` line 31 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7lBYs-_kBGrkFrU2N:-Oz7lBYs-_kBGrkFrU2O:bjvmw2k) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.cpp#L31)):*
> ```cpp
> namespace {
> 
> // Service names come from config, so they only need Prometheus quoting; track
> ```

Do other handlers do this and if so should there be a metrics utility header for it?
___
*[`src/admin/TrackMetricsHandler.cpp` line 68 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7lTmD1UPOoOVW30aJ:-Oz7lTmE33GW7tKPiugE:b-dua6j) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.cpp#L68)):*
> ```cpp
> };
> 
> std::unique_ptr<folly::IOBuf> formatPrometheus(const MoqxRelayContext::TrackMetricsResult& result) {
> ```

Same comment here about the helper?
___
*[`src/admin/TrackMetricsHandler.cpp` line 193 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7levfE6I56N-Fn9d9:-Oz7levfE6I56N-Fn9dA:b-nh8c0a) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.cpp#L193)):*
> ```cpp
> }
> 
> void sendError(proxygen::ResponseHandler* downstream, int status, const std::string& message) {
> ```

This almost certainly belongs in a helper
___
*[`src/admin/TrackMetricsHandler.cpp` line 237 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7lkBOEnQ0ZDOW5Q7e:-Oz7lkBOEnQ0ZDOW5Q7f:b-905l20) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.cpp#L237)):*
> ```cpp
>         auto nsPrefix = parseSafeNamespace(nsParam);
>         if (!nsPrefix) {
>           sendError(downstream, 400, "namespace is not in the MoQT safe name form\n");
> ```

MOQT not MoQT
___
*[`src/config/ConfigResolver.cpp` line 1204 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7lviODX2VkmC20Byg:-Oz7lviODX2VkmC20Byh:bz6ygsc) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/config/ConfigResolver.cpp#L1204)):*
> ```cpp
>     }
>     if (auto limit = adminOptional->track_metrics_limit.value()) {
>       resolved.trackMetricsLimit = *limit;
> ```

Need a verifier that default is < max
___
*[`src/relay/TrackStatsFilter.cpp` line 106 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7mInx73M6Lrb4Bkbe:-Oz7mInx73M6Lrb4Bkbf:b62pth1) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/relay/TrackStatsFilter.cpp#L106)):*
> ```cpp
> TrackStatsFilter::~TrackStatsFilter() {
>   bool egress = direction_ == Direction::Egress;
>   if (stats_->onOwnerThread()) {
> ```

What are the scenarios where this is destroyed in the wrong thread?  Can we rewrite with a lambda and if (onOnwer) fn else exec->add(fn)
___
*[`src/stats/TrackStats.h` line 33 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7mvhe8_TklzItrYeZ:-Oz7mvhe8_TklzItrYe_:blipfef) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/stats/TrackStats.h#L33)):*
> ```objc
> // so every field must have a meaningful merge.
> struct TrackCounters {
>   uint64_t groupsReceived{0};
> ```

Should we make a single struct of these counter values and hold in received and sent?  I think that could alleviate some conditionals we have above?
___
*[`src/stats/TrackStats.h` line 178 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7n_Yy7JMMjyWTgSPg:-Oz7n_Yy7JMMjyWTgSPh:b-okcaue) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/stats/TrackStats.h#L178)):*
> ```objc
>   friend class TrackStats;
> 
>   struct Slot {
> ```

Slot seems like an undescriptive name, pick a better one
___
*[`src/stats/TrackStats.h` line 204 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7nx0C4ovTZxn4y6De:-Oz7nx0C4ovTZxn4y6Df:bs2l6rn) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/stats/TrackStats.h#L204)):*
> ```objc
> };
> 
> inline folly::Executor* TrackStats::owningExec() const {
> ```

If there was a way to guarantee destruction on the right thread, I find it preferable to tracking this.  
___
*[`src/stats/TrackStatsRegistry.h` line 36 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7oNeE5FkzpRPEuQlQ:-Oz7oNeE5FkzpRPEuQlR:b5zwpc0) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/stats/TrackStatsRegistry.h#L36)):*
> ```objc
>   // For callers already running on the executor's thread, where dispatching
>   // and waiting would deadlock.
>   void bindHere(folly::Executor* exec);
> ```

This seems to only be used in tests -- is there a way to use bindAll instead and remove this API, merge addCollector into bindAll and remove the dup check?

afrind (commented):
> @afrind made 16 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)* status: 4 of 32 files reviewed, 1 unresolved discussion (waiting on gmarzot).

___
*[`docs/config.md` line 436 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7hX_rCXwsMyxGvOXO:-Oz9hbfB3TumgqMS5HXU:b-hff6wv) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/config.md#L436)):*

track_metrics_endpoint_default_limit
</blockquote></details>

renamed
___
*[`docs/config.md` line 437 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7hbAx4xpqTnUmsbBW:-Oz9hdlJ9q_Mi_88FuNQ:b-hff6wv) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/config.md#L437)):*

track_metrics_endpoint_max_limit
</blockquote></details>

renamed
___
*[`docs/metrics.md` line 222 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7hvuz6g13qJL_VtTH:-Oz9i9aB88IdSqb1AJVR:b-hfpv0l) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/docs/metrics.md#L222)):*

Remove the "Because the encoding is injective, " bit
</blockquote></details>

removed
___
*[`src/MoqxRelayContext.cpp` line 177 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7jMlcBE-sRFzup1yI:-Oz9iVqpBoXGyrFPNNtu:bb8njyr) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/MoqxRelayContext.cpp#L177)):*

call this initTrackStats
</blockquote></details>

Oops, I meant the boolean param should be "initTrackStats" - the fn should still be initThreadStatsCollectors
___
*[`src/MoqxRelayContext.cpp` line 338 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7k7--E0kzVJe177vN:-Oz9gs-G8s1fmpdyHi1u:bophx0j) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/MoqxRelayContext.cpp#L338)):*

Why are we scanning keys here instead of just iterating counters and pushing back service it->first, it->second?
</blockquote></details>

removed key scan
___
*[`src/admin/TrackMetricsHandler.h` line 19 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7kptE12h6OaiWQ4WU:-Oz9glqpAmNCvc-yKGD3:b-wjtjrx) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.h#L19)):*

this is sort of the same as the conig struct?  Is it a normal pattern to include it here?
</blockquote></details>

decided to keep the struct
___
*[`src/admin/TrackMetricsHandler.cpp` line 31 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7lBYs-_kBGrkFrU2N:-Oz9gwht8z5MeJfJlSQs:b-qjbqtk) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.cpp#L31)):*

Do other handlers do this and if so should there be a metrics utility header for it?
</blockquote></details>

Moved to utilties
___
*[`src/admin/TrackMetricsHandler.cpp` line 68 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7lTmD1UPOoOVW30aJ:-Oz9h-uW-FNd7VsIvX1J:b-ewjxen) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.cpp#L68)):*

Same comment here about the helper?
</blockquote></details>

moved to helper
___
*[`src/admin/TrackMetricsHandler.cpp` line 193 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7levfE6I56N-Fn9d9:-Oz9hBMr5rN0fGhApbZp:bbn5mle) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.cpp#L193)):*

This almost certainly belongs in a helper
</blockquote></details>

Added PrometheusWriter
___
*[`src/admin/TrackMetricsHandler.cpp` line 237 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7lkBOEnQ0ZDOW5Q7e:-Oz9hGGD57hoq0FQTIOY:b5x0d2h) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/admin/TrackMetricsHandler.cpp#L237)):*

MOQT not MoQT
</blockquote></details>

Fixed - MOQT
___
*[`src/config/ConfigResolver.cpp` line 1204 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7lviODX2VkmC20Byg:-Oz9hKMk3Qs8TYj1gYvH:b-hlv2ot) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/config/ConfigResolver.cpp#L1204)):*

Need a verifier that default is < max
</blockquote></details>

Added check
___
*[`src/relay/TrackStatsFilter.cpp` line 106 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7mInx73M6Lrb4Bkbe:-Oz9hNnR5wXm-as19jxC:b-wg7sus) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/relay/TrackStatsFilter.cpp#L106)):*

What are the scenarios where this is destroyed in the wrong thread?  Can we rewrite with a lambda and if (onOnwer) fn else exec->add(fn)
</blockquote></details>

Used a lambda, but decided not to chase the cross-threadness
___
*[`src/stats/TrackStats.h` line 33 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7mvhe8_TklzItrYeZ:-Oz9hfzw1oQ8qUTgYC5T:b5a22s1) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/stats/TrackStats.h#L33)):*

Should we make a single struct of these counter values and hold in received and sent?  I think that could alleviate some conditionals we have above?
</blockquote></details>

Added DirectionCounters
___
*[`src/stats/TrackStats.h` line 178 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7n_Yy7JMMjyWTgSPg:-Oz9huA7ECGJbm7cPrWm:b-l0n4li) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/stats/TrackStats.h#L178)):*

Slot seems like an undescriptive name, pick a better one
</blockquote></details>

Slot -> TrackRef
___
*[`src/stats/TrackStats.h` line 204 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7nx0C4ovTZxn4y6De:-Oz9hwkJ38-IyvxCumKC:bh28kvi) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/stats/TrackStats.h#L204)):*

If there was a way to guarantee destruction on the right thread, I find it preferable to tracking this.
</blockquote></details>

Punted for now
___
*[`src/stats/TrackStatsRegistry.h` line 36 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7oNeE5FkzpRPEuQlQ:-Oz9i4EI8wQbOionCHfv:b-h7vu72) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/stats/TrackStatsRegistry.h#L36)):*

This seems to only be used in tests -- is there a way to use bindAll instead and remove this API, merge addCollector into bindAll and remove the dup check?
</blockquote></details>

removed "bindHere" by scoping the tests

afrind (commented):
> @afrind made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)* status: 4 of 32 files reviewed, all discussions resolved (waiting on gmarzot).

___
*[`src/MoqxRelayContext.cpp` line 177 at r2](https://reviewable.io/reviews/openmoq/moqx/533#-Oz7jMlcBE-sRFzup1yI:-Oz9keYy46pIo2m43dSK:b-7ql1td) ([raw file](https://github.com/openmoq/moqx/blob/dbbc46620e63d1f4280467c96c01c6cc570314b5/src/MoqxRelayContext.cpp#L177)):*

Oops, I meant the boolean param should be "initTrackStats" - the fn should still be initThreadStatsCollectors
</blockquote></details>

fixed

michalhosna (changes requested):
> @michalhosna made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)* status: 4 of 32 files reviewed, 2 unresolved discussions (waiting on afrind and gmarzot).

___
*[`docs/config.md` line 451 at r3](https://reviewable.io/reviews/openmoq/moqx/533#-OzHp-8V0Y7NULfuDdUV:-OzHp-8V0Y7NULfuDdUW:b-pv5zlo) ([raw file](https://github.com/openmoq/moqx/blob/e6e55ef46d4c2d39537b5fd3fa331161d5603925/docs/config.md#L451)):*
> ```Markdown
> guard rail, not a selection rule: a query matching more tracks than the limit is
> rejected rather than truncated, because an arbitrary subset would give
> Prometheus a series set that reshuffles between scrapes.
> ```

Nit: I think this documentation is relatively confusing without knowing the query structure. Maybe point to the full docs? 
___
*[`src/admin/TrackMetricsHandler.cpp` line 215 at r3](https://reviewable.io/reviews/openmoq/moqx/533#-OzHutXuDxJDUppQESq0:-OzHutXvATgnUt7X5ojF:b-ufzb9c) ([raw file](https://github.com/openmoq/moqx/blob/e6e55ef46d4c2d39537b5fd3fa331161d5603925/src/admin/TrackMetricsHandler.cpp#L215)):*
> ```cpp
>             return;
>           }
>           limit = std::min(*parsed, limits.maxLimit);
> ```

If maxLimit ==0, this gets set to `0` i.e. always returns 400? Is that correct? 
Why is there the `std::max<size_t>(1,` above? What's the reason of `1` floor there? 

Should `maxLimit = 0` have special meaning? Like unlimited?
___
*[`src/config/Config.h` line 230 at r3](https://reviewable.io/reviews/openmoq/moqx/533#-OzHrik94TLi3u5bVeuN:-OzHrik94TLi3u5bVeuO:bplzopy) ([raw file](https://github.com/openmoq/moqx/blob/e6e55ef46d4c2d39537b5fd3fa331161d5603925/src/config/Config.h#L230)):*
> ```objc
>   folly::SocketAddress address;
>   std::optional<TlsConfig> tls;
>   bool trackMetricsEnabled{true};
> ```

Why `true` by default? Isn't the collection somewhat expensive? 

If you do not know about this feature, it seems much more dangerous to leave it enabled than useful to leave it enabled. If you do know about this feature, you don't care about default. 

Or, as I like explicit over implicit, require decision in config file and have no default, but that's probably too harsh here

afrind (commented):
> @afrind made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)* status: 4 of 32 files reviewed, 2 unresolved discussions (waiting on gmarzot and michalhosna).

___
*[`docs/config.md` line 451 at r3](https://reviewable.io/reviews/openmoq/moqx/533#-OzHp-8V0Y7NULfuDdUV:-OzI5A5M11CT3CCaBPjv:b32wtd) ([raw file](https://github.com/openmoq/moqx/blob/e6e55ef46d4c2d39537b5fd3fa331161d5603925/docs/config.md#L451)):*

Nit: I think this documentation is relatively confusing without knowing the query structure. Maybe point to the full docs?
</blockquote></details>

ok
___
*[`src/admin/TrackMetricsHandler.cpp` line 215 at r3](https://reviewable.io/reviews/openmoq/moqx/533#-OzHutXuDxJDUppQESq0:-OzI4uNJ2nQPPrWerNr3:bqahqx3) ([raw file](https://github.com/openmoq/moqx/blob/e6e55ef46d4c2d39537b5fd3fa331161d5603925/src/admin/TrackMetricsHandler.cpp#L215)):*

If maxLimit ==0, this gets set to `0` i.e. always returns 400? Is that correct? 
Why is there the `std::max<size_t>(1,` above? What's the reason of `1` floor there? 

Should `maxLimit = 0` have special meaning? Like unlimited?
</blockquote></details>

I don't think we want to allow unlimited.  I'll add a config check that maxLimit > 0 when enabled=True
___
*[`src/config/Config.h` line 230 at r3](https://reviewable.io/reviews/openmoq/moqx/533#-OzHrik94TLi3u5bVeuN:-OzI51jCEnn3Y9NXLYa3:b-6qmzmz) ([raw file](https://github.com/openmoq/moqx/blob/e6e55ef46d4c2d39537b5fd3fa331161d5603925/src/config/Config.h#L230)):*

Why `true` by default? Isn't the collection somewhat expensive? 

If you do not know about this feature, it seems much more dangerous to leave it enabled than useful to leave it enabled. If you do know about this feature, you don't care about default. 

Or, as I like explicit over implicit, require decision in config file and have no default, but that's probably too harsh here
</blockquote></details>

Sure we can default to false.

afrind (commented):
> @afrind made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)* status: 3 of 33 files reviewed, 1 unresolved discussion (waiting on gmarzot and michalhosna).

___
*[`src/admin/TrackMetricsHandler.cpp` line 215 at r3](https://reviewable.io/reviews/openmoq/moqx/533#-OzHutXuDxJDUppQESq0:-OzIA3h37HbOLNHhPBzi:bgwgh1s) ([raw file](https://github.com/openmoq/moqx/blob/e6e55ef46d4c2d39537b5fd3fa331161d5603925/src/admin/TrackMetricsHandler.cpp#L215)):*

I don't think we want to allow unlimited.  I'll add a config check that maxLimit > 0 when enabled=True
</blockquote></details>

I removed the floor of 1, and 400 anyone who requests > max.

michalhosna (commented):
> @michalhosna resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)* status: 3 of 33 files reviewed, all discussions resolved (waiting on gmarzot).

gmarzot (approved):
> @gmarzot reviewed 31 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

michalhosna (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/533#-:-OzhFMZ220-1P6z2xzFH:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #541, opened 2026-08-04
relay: name the initial track state a subscriber must observe

afrind:
A track's starting largest location and extensions were passed around as two loose values and applied by hand at five sites, in two different orders.

Group them as InitialTrackState. capture() reads the pair off a live forwarder; applyTo() writes it to a forwarder or to a subscriber, whichever the caller has. The two fields are independent, so the order applyTo picks is arbitrary. No behavior change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/541)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -9,6 +9,7 @@
 #include "MoqxRelay.h"
 #include "relay/CrossExecFilter.h"
 #include "relay/CrossExecForwarderCallback.h"
+#include "relay/InitialTrackState.h"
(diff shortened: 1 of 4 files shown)


akash-a-n (approved):
> @akash-a-n reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/541)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #542, opened 2026-08-05
relay: give LocalForwarderRegistry a three-state entry

afrind:
The registry mapped a track name straight to a forwarder, which let a reader attach to one before its largest and extensions arrived from upstream.

An entry is now absent, pending, or ready, and a pending entry does not hand out its forwarder. Only the Claim that join() returns can reach it, and dropping that Claim fails the entry instead of stranding whoever waits on it.

Nothing observes the pending state yet, because both callers mark the entry ready on the next line; holding a claim open across setup is what the commits above this one do. LocalForwarderRegistryTest is new, and covers the three states, the identity checks, and the two asserts a caller can trip.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/542)


akash-a-n (approved):
> @akash-a-n reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/542)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #545, opened 2026-08-05
relay: make subscribers wait for an in-flight local forwarder setup

afrind:
A subscriber that arrived while another subscriber's upstream subscribe was still in flight joined the local forwarder right away: it got a SUBSCRIBE_OK with no largest object, and it kept its subscription even when the upstream subscribe came back with an error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/545)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -1188,15 +1188,19 @@ folly::coro::Task<void> MoqxRelay::addSubscriberAndPublishViaLocalForwarder(
   co_await awaitPublishReply(localFwd, std::move(p->subscriber), std::move(p->reply));
 }
 
+LocalForwarderRegistry& MoqxRelay::localRegistry() {
+  if (!tlForwarders_.get()) {
+    tlForwarders_.reset(new LocalForwarderRegistry());
+  }
+  return *tlForwarders_;
+}
+
 // Slow-path local-forwarder bootstrap shared by the publish and subscribe LF paths.
 // Callers handle the fast path and install the PendingForwarderCallback themselves,
 // because the timing differs.
 MoqxRelay::LocalForwarderBootstrap
 MoqxRelay::acquireLocalForwarder(const FullTrackName& ftn, const InitialTrackState& initial) {
-  if (!tlForwarders_.get()) {
-    tlForwarders_.reset(new LocalForwarderRegistry());
-  }
-  auto* localReg = tlForwarders_.get();
+  auto* localReg = &localRegistry();
   auto joined = localReg->join(ftn, [&] { return std::make_shared<MoQForwarder>(ftn); });
   if (auto* claim = std::get_if<LocalForwarderRegistry::Claim>(&joined)) {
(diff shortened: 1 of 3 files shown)


akash-a-n (approved):
> @akash-a-n reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/545)* status: all files reviewed, 1 unresolved discussion (waiting on afrind).

___
*[`src/MoqxRelay.cpp` line 1950 at r1](https://reviewable.io/reviews/openmoq/moqx/545#-OzR0STX9Kid8pF4bSE1:-OzR0STX9Kid8pF4bSE2:b-s818cq) ([raw file](https://github.com/openmoq/moqx/blob/b38c6c2bf2debe2a2e84cd5f8d0b245cbecd3b86/src/MoqxRelay.cpp#L1950)):*
> ```cpp
>     co_await folly::coro::co_awaitTry(std::move(pending->ready));
>     // Re-resolve: the entry may have been displaced while we were waiting. A failed setup
>     // fails this subscriber too.
> ```

I don't think I completely understand this part. Is it enough to just use getIfReady? shouldn't there be a co-await or a loop of sorts?

afrind (commented):
> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/545)* status: all files reviewed, 1 unresolved discussion (waiting on akash-a-n).

___
*[`src/MoqxRelay.cpp` line 1950 at r1](https://reviewable.io/reviews/openmoq/moqx/545#-OzR0STX9Kid8pF4bSE1:-OzYHjm51pQ1iWNV7r_l:bmqgzy8) ([raw file](https://github.com/openmoq/moqx/blob/b38c6c2bf2debe2a2e84cd5f8d0b245cbecd3b86/src/MoqxRelay.cpp#L1950)):*

I don't think I completely understand this part. Is it enough to just use getIfReady? shouldn't there be a co-await or a loop of sorts?
</blockquote></details>

We're getting into complex territory here.  I think the case is:

1st Subscriber in this thread arrives, gets an entry in LF registry, begins setup
2nd Subscriber (this one) arrives, waits on pending

a) A publisher arrives and does a PUBLISH; this will overwrite the entry.  OR
b) The original subscribe setup failed

If it's a) getIfReady returns ready that points to the new publisher's forwarder
If it's b) getIfReady returns !ready, and we fail the waiting subscriber

Of course there's the happy path which is that setup succeeded and the 1st subscriber called markReady.  Then ready will be the same one we were waiting on.

There might be a ways to continue to improve this (eg a subscriber that gets a failed setup could try again), but it's easy to get lost here, and I want to keep the changes small.

akash-a-n (approved):
> @akash-a-n resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/545)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #546, opened 2026-08-05
relay: install the local forwarder chain on subscribe-created tracks

afrind:
A track created by its first subscriber left no entry in the publisher thread's LocalForwarderRegistry, so a subscriber landing on that thread joined a separate local forwarder instead of the publisher's. First subscriber setup now calls installPublisherForwarder on the publisher executor with removeOnEmpty=true, so the entry is there to find and is dropped with the last subscriber.

The initial state for the requesting subscriber is captured from the publisher forwarder, so subscribers joining an existing attachment also get the most recent largest and extensions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/546)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -458,20 +458,18 @@ MoqxRelay::validatePublishNamespace(const FullTrackName& ftn, RequestID requestI
 // source ends, so a chain over a forwarder that does not hold it has nothing to remove.
 LocalForwarderRegistry::Claim MoqxRelay::installPublisherForwarder(
     const FullTrackName& ftn,
-    const std::shared_ptr<MoQForwarder>& fwd
+    const std::shared_ptr<MoQForwarder>& fwd,
+    bool removeOnEmpty
 ) {
-  // removeOnEmpty=false: the publisher's forwarder must survive subscriber churn.
+  auto& localReg = localRegistry();
   auto relayAdapter = std::make_shared<WeakRelayForwarderCallback>(weak_from_this());
   auto crossExec =
       std::make_shared<CrossExecForwarderCallback>(relayExec_, fwd, std::move(relayAdapter));
-  fwd->setCallback(std::make_shared<LocalForwarderCallback>(
-      tlForwarders_.get(),
-      ftn,
-      std::move(crossExec),
-      /*removeOnEmpty=*/false
-  ));
+  fwd->setCallback(
+      std::make_shared<LocalForwarderCallback>(&localReg, ftn, std::move(crossExec), removeOnEmpty)
+  );
 
-  return tlForwarders_->replace(ftn, fwd);
+  return localReg.replace(ftn, fwd);
 }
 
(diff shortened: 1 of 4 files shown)


akash-a-n (changes requested):
> @akash-a-n reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/546)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on afrind).

___
*[`src/MoqxRelay.cpp` line 1198 at r1](https://reviewable.io/reviews/openmoq/moqx/546#-Ozh1NX-5o9kLRoKjRE5:-Ozh1NX-5o9kLRoKjRE6:bahz8h2) ([raw file](https://github.com/openmoq/moqx/blob/2ec50efd18e6adb047f7af1e657902d7586faca6/src/MoqxRelay.cpp#L1198)):*
> ```cpp
>   }
>   auto* ready = std::get_if<LocalForwarderRegistry::Ready>(&joined);
>   XCHECK(ready) << "local forwarder entry still pending; a caller failed to resolve its claim: "
> ```

copilot flagged a race condition here, where the joinResult would be pending:
1. On thread T we have an open subscribe_namespace request
2. We get a SUBSCRIBE request on the same thread which then goes upstream and gets suspended
3. During this suspension, we get a publish(on a different thread) that fans out thread T through acquireLocalForwarder and the joinResult would be 'pending'

afrind (commented):
> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/546)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on akash-a-n).

___
*[`src/MoqxRelay.cpp` line 1198 at r1](https://reviewable.io/reviews/openmoq/moqx/546#-Ozh1NX-5o9kLRoKjRE5:-OzhXuLq3PTuk7__kInu:b-1ij547) ([raw file](https://github.com/openmoq/moqx/blob/2ec50efd18e6adb047f7af1e657902d7586faca6/src/MoqxRelay.cpp#L1198)):*

copilot flagged a race condition here, where the joinResult would be pending:
1. On thread T we have an open subscribe_namespace request
2. We get a SUBSCRIBE request on the same thread which then goes upstream and gets suspended
3. During this suspension, we get a publish(on a different thread) that fans out thread T through acquireLocalForwarder and the joinResult would be 'pending'
</blockquote></details>

Aha, good find.  I'm going to handle this by dropping a PUBLISH that matches an in-progress SUBSCRIBE.  I think it's important because the subscription should get the `pinned` flag.

akash-a-n (approved):
> @akash-a-n reviewed 3 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/546)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

Issue #530, opened 2026-07-30
Auth: anonymous claim


michalhosna:
Today a service is either authorization-less or token-required. 

The only exception is `require_setup_token: false`

There's a use case for allowing specific operations token-less on an otherwise token-protected service. e.g. anonymous subscribe/fetch for anyone, while publishing still requires a valid token.

Defining what's allowed may be complex. For CAT-based auth, I propose an "anonymous claim": a statically-defined [moqt claim](https://moq-wg.github.io/CAT-4-MOQT/draft-ietf-moq-c4m.html#section-2.1) in the config that is applied to requests without a token.

Details TODO. e.g. should this apply only to requests with no token, or also to requests whose token is invalid? And is "valid" meant as decidable, or as claim-satisfying?

michalhosna:
I am going to add support for multiple auth tokens in general (see https://github.com/moq-wg/moq-transport/issues/1838). 

And then treat the anonymous claim as another token in the message. I.e. what allows this anonymous claim will always be allowed for everyone regardles of what tokens they send.

PR #552, opened 2026-08-07
auth: support multiple AUTHORIZATION_TOKEN in single message

michalhosna:
Preparation for #530 

Related:
- https://github.com/moq-wg/moq-transport/issues/1838

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/552)

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -179,27 +179,43 @@ folly::Expected<Grants, AuthError> AuthTokenVerifier::verify(const AuthToken& to
   return folly::makeUnexpected(AuthError::BadSignature);
 }
 
-folly::Expected<std::shared_ptr<const Grants>, AuthError>
+folly::Expected<std::shared_ptr<const std::vector<Grants>>, AuthError>
 authenticateSetup(const AuthTokenVerifier& verifier, const Parameters& setupParams) {
   if (!verifier.enabled()) {
-    return std::shared_ptr<const Grants>{};
+    return std::shared_ptr<const std::vector<Grants>>{};
   }
-  if (auto token = findAuthToken(setupParams, verifier.tokenType())) {
-    auto verified = verifier.verify(*token);
+
+  auto tokens = findAuthTokens(setupParams, verifier.tokenType());
+  std::vector<Grants> verifiedGrants;
+  std::optional<AuthError> firstError;
+  for (const auto& token : tokens) {
+    auto verified = verifier.verify(token);
     if (verified.hasError()) {
-      return folly::makeUnexpected(verified.error());
+      if (!firstError) {
+        firstError = verified.error();
+      }
+      continue;
     }
-    if (!allows(verified.value(), Action::ClientSetup, TrackNamespace{})) {
-      return folly::makeUnexpected(AuthError::Forbidden);
-    }
-    return std::make_shared<const Grants>(std::move(verified.value()));
+    verifiedGrants.push_back(std::move(verified.value()));
   }
+
+  if (!tokens.empty()) {
+    // At least one setup token was presented; the session is authorized iff
+    // any of the verified ones grants ClientSetup. All verified tokens' other
+    // scopes still pool into the session grants below, not just the one that
+    // happened to grant ClientSetup.
+    if (!allowsAny(verifiedGrants, Action::ClientSetup, TrackNamespace{})) {
+      return folly::makeUnexpected(firstError.value_or(AuthError::Forbidden));
+    }
+    return std::make_shared<const std::vector<Grants>>(std::move(verifiedGrants));
+  }
+
   if (verifier.requireSetupToken()) {
     return folly::makeUnexpected(AuthError::Missing);
   }
-  // No setup token but not required: connect with empty grants; requests must
-  // then carry their own tokens to be authorized.
-  return std::make_shared<const Grants>();
+  // No setup token but not required: connect with empty session grants;
+  // requests must then carry their own tokens to be authorized (see authorize()).
+  return std::make_shared<const std::vector<Grants>>();
 }
 
 folly::Expected<folly::Unit, AuthError> authorize(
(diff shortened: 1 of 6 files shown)


mondain:
> _[Reviewable](https://reviewable.io/reviews/openmoq/moqx/552#-)_

I found two blocking authorization-contract issues.

[IMPORTANT] Invalid tokens must not be dropped when another credential permits the operation. `authenticateSetup()` and `authorize()` currently continue after every verification error and accept the message if another token or the session grants authorize it. Section 9.2.2.1 of the checked-in draft-16 says a message containing a well-formed Token structure but otherwise invalid AUTHORIZATION TOKEN parameter MUST be rejected with `MALFORMED_AUTH_TOKEN`. The existing `AuthorizeGarbageRequestTokenDoesNotBlockSessionGrants` and `AuthenticateSetupAcceptsAnyOfMultipleSetupTokens` tests codify the opposite behavior. I added a temporary diagnostic with a bad-signature request token followed by a valid token; the expected rejection failed because `authorize()` returned success. The linked WG issue #1838 is still open and explicitly identifies this conflict, so the current draft behavior should remain authoritative until that discussion changes it. Please fail on a recognized token's verification error and add negative request/setup coverage; union grants only across tokens that are all valid.

[IMPORTANT] A request token no longer overrides the session grants. `allow_request_token_override` is documented in `docs/config.md`, `config.example.yaml`, and the generated schema as replacing the setup grants for that request. The new `requestGrants || sessionGrants` check makes the sources additive instead. That means a request token intended to narrow authority cannot do so: a broad session grant still authorizes the operation. A temporary diagnostic with a valid Publish-only request token and a Subscribe session grant expected `Forbidden`, but returned success. When override is enabled and at least one request token is present, please evaluate the request-token pool instead of the session pool; use session grants only when the request carries no matching token. If additive behavior is intentional, the setting and all three operator-facing descriptions need an explicit contract change.

For verification, exact head `138350f6` built successfully, all 36 focused auth/filter/context tests passed before the diagnostics, and `git diff --check` is clean.

michalhosna:
@mondain I am lost at what reaction/action do you expect from me? 

Both of those findings work as I intended, at least what I can parse from your text, do you disagree with my intention or are you raising a question if this was the intention?

afrind (changes requested):
> The linked WG issue #1838 is still open and explicitly identifies this conflict, so the current draft behavior should remain authoritative until that discussion changes it.

eh, I think we have operational flexibility here, especially if we think it's going to trend towards this interpretation.

> A request token no longer overrides the session grants.

As long as the docs are updated to match, I think this is also within our power to define.  I think "SETUP tokens are the base set, requests extend privileges" makes sense.  Obviously the caller _has_ the SETUP token so they could have included it in the request if they wanted to.

> @afrind reviewed 6 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/552#-:-OzhD-t07B96A7PlKoaa:b-9oewfs)* status: all files reviewed, 2 unresolved discussions (waiting on michalhosna).

___
*[`src/auth/Auth.h` line 131 at r1](https://reviewable.io/reviews/openmoq/moqx/552#-OzhC1EcAIIMeadINfZm:-OzhC1EcAIIMeadINfZn:b-cak5bp) ([raw file](https://github.com/openmoq/moqx/blob/138350f6e40da9ab490e97be5f1e79d55456b648/src/auth/Auth.h#L131)):*
> ```objc
> 
> // Verifies the setup AUTHORIZATION_TOKEN(s). Returns a null pointer when auth
> // is disabled; otherwise a shared vector of every successfully-verified
> ```

why shared and not unique_ptr?
___
*[`src/auth/Auth.cpp` line 182 at r1](https://reviewable.io/reviews/openmoq/moqx/552#-OzhCKVj5v4tJckBNBNo:-OzhCKVj5v4tJckBNBNp:bo7ct0s) ([raw file](https://github.com/openmoq/moqx/blob/138350f6e40da9ab490e97be5f1e79d55456b648/src/auth/Auth.cpp#L182)):*
> ```cpp
> }
> 
> folly::Expected<std::shared_ptr<const std::vector<Grants>>, AuthError>
> ```

I guess I missed before that this is shared_ptr - can it be unique?

michalhosna (commented):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/552)* status: all files reviewed, 2 unresolved discussions (waiting on afrind).

___
*[`src/auth/Auth.h` line 131 at r1](https://reviewable.io/reviews/openmoq/moqx/552#-OzhC1EcAIIMeadINfZm:-Ozki4Ts5-7rKVCKNFg2:b-ebhvd1) ([raw file](https://github.com/openmoq/moqx/blob/138350f6e40da9ab490e97be5f1e79d55456b648/src/auth/Auth.h#L131)):*

why shared and not unique_ptr?
</blockquote></details>

It's passed down to `AuthPublisherFilter` and `AuthSubscriberFilter`. 

AFAIK the Filter's lifetime is decoupled from Sessions's lifetime, so making it owned by the session is not safe as the filter may outlive it in coroutine. 

And I don't feel like single shared_ptr copy at session initialization is worth optimizing.

afrind (commented):
> @afrind made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/552)* status: all files reviewed, 1 unresolved discussion (waiting on michalhosna).

___
*[`src/auth/Auth.h` line 131 at r1](https://reviewable.io/reviews/openmoq/moqx/552#-OzhC1EcAIIMeadINfZm:-OzlQx3M7YCKDKIRoLqd:b-k67ijj) ([raw file](https://github.com/openmoq/moqx/blob/138350f6e40da9ab490e97be5f1e79d55456b648/src/auth/Auth.h#L131)):*

It's passed down to `AuthPublisherFilter` and `AuthSubscriberFilter`. 

AFAIK the Filter's lifetime is decoupled from Sessions's lifetime, so making it owned by the session is not safe as the filter may outlive it in coroutine. 

And I don't feel like single shared_ptr copy at session initialization is worth optimizing.
</blockquote></details>

I think it's less about optimization and more about making it easier to reason about things that have only a single owner.  But I won't lie down in the road here.

afrind (approved):
> @afrind resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/552)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna).

PR #561, opened 2026-08-12
admin: honor the ?limit= parameter on /metrics/track

### LINKED ISSUE #560: /metrics/track ignores the limit query parameter
`limit` has no effect; the endpoint always uses `admin.track_metrics_endpoint_default_limit`.

```
$ curl -s 'localhost:8000/metrics/track?limit=1000'
matched 17 tracks, exceeds limit=10; narrow the namespace or raise limit

$ curl -s 'localhost:8000/metrics/track?limit=50'
matched 17 tracks, exceeds limit=10; narrow the namespace or raise limit
```

`docs/metrics.md` documents it as accepted and clamped to `track_metrics_endpoint_max_limit` (1000), and the error message tells the caller to raise a limit that raising does not change.

Namespace scoping works, so per-namespace scrapes are unaffected (#539 relies on that). It blocks the documented escape hatch for a namespace holding more tracks than the default.

Seen on v0.2.1-48-g3ba4740e.

afrind:
The parsed value was validated and discarded, so every scrape used admin.track_metrics_endpoint_default_limit and the 400 body told callers to raise a limit that raising did not change.

Also drop the doc's claim that an over-max limit is clamped; it is a 400.

Fixes #560

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/561)

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -200,7 +200,7 @@ goes away, and start from zero if it comes back.
 | `namespace` | no | Namespace prefix in the safe form below. Matches every track under the prefix unless `track` is given. Default: all namespaces. |
 | `service` | no | Restrict to one service. Default: all services, each labeled. |
 | `track` | no | Exact track name within the namespace. |
-| `limit` | no | Max tracks to report. Default `admin.track_metrics_endpoint_default_limit` (10), clamped to `admin.track_metrics_endpoint_max_limit` (1000). |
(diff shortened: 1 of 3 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/561)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gmarzot).

PR #573, opened 2026-08-13
relay: send track name across executors, not forwarder refs

afrind:
CrossExecForwarderCallback locked its weak_ptr on the source thread and moved the resulting shared_ptr into the lambda it posted to the target executor. folly destroys that lambda after invoking it, on the target thread, so when that ref was the last one ~MoQForwarder ran off its owner. The destructor detaches subgroups, which resets QUIC streams and trips EventBase's evbTid == curTid check (moqx-io0 vs moqx-relay0). Under ASan, relay_chain aborted during teardown on the first iteration.

No consumer needed the forwarder: every downstream either read only fullTrackName() or ignored the pointer entirely, and onPublishDone reached a no-op base method. Read the name off the forwarder that invoked the callback, on its owner thread where it is alive, and send only that; recipients re-resolve by name.

Adds TrackEventCallback as the name-based downstream interface. WeakRelayForwarderCallback implements both it and MoQForwarder::Callback, since SingleThread/RelayExec install it directly on the forwarder with no hop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/573)

PR #574, opened 2026-08-13
relay: add ForwarderRef, an owner-thread handle to a MoQForwarder

afrind:
A MoQForwarder belongs to the executor that created it, but we hand shared_ptrs to it across threads, where a holder can dereference it or run its destructor off that executor. ForwarderRef can be held anywhere and only used on the owner: post() and co_with() send work there, and pinOnOwner() gives scoped strong access that debug-checks the calling thread. There is no get() and no operator->, so there is no way to reach the forwarder from the wrong thread. identityKey() names a forwarder by its address plus a generation the type mints itself, which keeps a dead publication nameable and stops a later forwarder that reuses the address from comparing equal to it.

Nothing uses it yet; the call sites land in subsequent commits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/574)

### DIFF
--- a/src/relay/ForwarderRef.h
+++ b/src/relay/ForwarderRef.h
@@ -1,0 +1,178 @@
+/*
+ * Copyright (c) OpenMOQ contributors.
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <moxygen/relay/MoQForwarder.h>
+
+#include <folly/Executor.h>
+#include <folly/coro/Task.h>
+#include <folly/logging/xlog.h>
+
+#include <memory>
+#include <optional>
+#include <type_traits>
+
+namespace openmoq::moqx {
+
+// A track and the executor its forwarder lives on.  The receiver of this ref
+// can hop back to exec to get the latest publisher for this track.
+struct TrackRef {
+  moxygen::FullTrackName ftn;
+  folly::Executor* exec{nullptr};
+};
+
+// A reference to a MoQForwarder that may be held on any thread.  Two modes, fixed
+// at construction:
+//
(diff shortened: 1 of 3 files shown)


michalhosna (commented):
This design seems weird to me / I am not sure if I get it. 

You claim "Unrepresentable" but it feel a lot like convention with at best debug checks. 

Like copy in owned mode seems unsafe? The modes creations are not guarded at all. 
pinOnOwner can be called on on-owners? Those feels like design decisions, not something that was overlooked. 

Shouldn't we just have strong move only type and weak copy type? Or templates? 

Also #573 was about not sharing refs across refs. But now we want to share refs? 

Generally, I don't feel like this is reviewable/understandable without seeing the envisioned use case, but being it PR stack pushes this in the direction that this should be independently understandable, but I don't get it (And neither agent I ask gets the current design). 

> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/574#-:-P-Jpsoa4emXWF2eLcgD:bsgfzk9)* status: 0 of 3 files reviewed, all discussions resolved.

michalhosna (approved):
:lgtm:

> @michalhosna reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/574#-:-P-Vg_t76hgrQm90ScY9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

michalhosna (approved):
> @michalhosna reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/574)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #576, opened 2026-08-13
relay: let LocalForwarderRegistry park a displaced forwarder

afrind:
replace() overwrites a registry entry and drops the last reference to the forwarder it displaced, so a caller that still needs the old one — usually to drain it — has no way to reach it. replaceAndPark() keeps that forwarder alive in a side map instead, and takeDisplaced() hands it back. This lets a decision about the old forwarder travel to another thread and come back, because the reference stays anchored on the owning thread and ~MoQForwarder cannot run off it. The park returns a move-only ParkTicket that takeDisplaced() consumes, so dropping the obligation aborts right where it was incurred instead of much later at iothread teardown, with nothing linking the abort to the call that stranded the forwarder.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/576)

### DIFF
--- a/src/relay/LocalForwarderRegistry.h
+++ b/src/relay/LocalForwarderRegistry.h
@@ -16,6 +16,7 @@
 #include <folly/logging/xlog.h>
 
 #include <variant>
+#include <vector>
 
 namespace openmoq::moqx {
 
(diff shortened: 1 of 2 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/576)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #577, opened 2026-08-13
relay: keep registry forwarders behind a ForwarderRef

afrind:
In local-forwarder mode a SubscriptionRegistry entry held a shared_ptr to a forwarder that lives on the publisher's executor, so the relay executor could read its counters, run its destructor, or hand it to a subscriber thread that later found it displaced. Entries and every view over them now hold a ForwarderRef: a caller that owns the forwarder pins it inline, and everyone else posts to its executor. Entry cleanup keys off the entry's ForwarderId instead of locking a weak_ptr, which used to decline to erase an entry exactly when its forwarder was already gone, leaving an unfulfilled promise that every later subscriber to that track waited on forever. The local-forwarder fan-out now carries a track name and executor and resolves the publisher forwarder on that executor at the point of use, so a subscriber whose hop lands late fails with "publisher forwarder gone" rather than attaching to a forwarder nobody feeds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/577)

### DIFF
--- a/docs/dev/local-forwarder-flow.md
+++ b/docs/dev/local-forwarder-flow.md
@@ -48,33 +48,61 @@ Executor legend used below:
 
 ## Publish
 
-A publisher's local forwarder **is** the publisher forwarder — the same object serves as the
-on-thread data-plane forwarder and as the registry's publisher entry (no second forwarder is
-constructed on `relayExec_`, unlike the `RelayExec`-mode branch in `publishWithSession`).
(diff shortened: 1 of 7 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/577)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

gmarzot:
image

PR #587, opened 2026-08-13
relay: drain cross-executor messages before destroying the servers

afrind:
Shutdown teardown bounces between the io threads and the relay threads several times: a subscriber going away notifies the publisher's forwarder, which notifies the relay, which unsubscribes upstream. Those hops are fire-and-forget lambdas holding raw executor pointers, and ~MoQServer frees each thread's MoQExecutor, so a lambda still queued when the servers went away called add() on freed memory.

drainExecs runs a barrier across the io EVBs and every relay executor, waiting for each round to finish everywhere before posting the next, so each round lets the cascade advance one more hop. main calls it after stopping the servers and before destroying them. This is a stopgap until those executors are held with KeepAlive tokens.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/587)

### DIFF
--- a/src/MoqxRelayContext.cpp
+++ b/src/MoqxRelayContext.cpp
@@ -16,6 +16,7 @@
 #include <folly/coro/Collect.h>
 #include <folly/coro/Task.h>
 #include <folly/executors/thread_factory/NamedThreadFactory.h>
+#include <folly/futures/Future.h>
 #include <folly/logging/xlog.h>
 
 using namespace moxygen;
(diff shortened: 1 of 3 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/587)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

gmarzot:
image

PR #590, opened 2026-08-13
test: fail integration tests when the relay exits non-zero

afrind:
Cleanup killed the relay and ignored its exit status, so a sanitizer-detected leak (ASan exits 1) or a hung teardown (the shutdown watchdog's _Exit) passed silently. Every test script that starts a relay now waits on it and fails if it exits non-zero. The waiting lives in test_relay_lifecycle.sh so the scripts don't each repeat it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/590)

### DIFF
--- a/test/test_admin_cache_purge.sh
+++ b/test/test_admin_cache_purge.sh
@@ -4,6 +4,8 @@ set -euo pipefail
 BINARY="${1:-$(dirname "$0")/../build/default/moqx}"
 # shellcheck source=test_ports.sh
 source "$(dirname "$0")/test_ports.sh"
+# shellcheck source=test_relay_lifecycle.sh
+source "$(dirname "$0")/test_relay_lifecycle.sh"
 LISTEN_PORT=$TEST_CACHE_PURGE_LISTEN
 ADMIN_PORT=$TEST_CACHE_PURGE_ADMIN
 INFO_URL="http://localhost:${ADMIN_PORT}/info"
@@ -17,11 +19,13 @@ fi
 TMPDIR=$(mktemp -d)
 MOQX_PID=""
 cleanup() {
+  local relay_failed=0
   if [[ -n "${MOQX_PID:-}" ]]; then
     kill "$MOQX_PID" 2>/dev/null || true
-    wait "$MOQX_PID" 2>/dev/null || true
+    reap_relays "$MOQX_PID" || relay_failed=1
   fi
   rm -rf "$TMPDIR"
+  (( relay_failed == 0 )) || exit 1
 }
 trap cleanup EXIT
 
--- a/test/test_admin_cache_purge_race.sh
+++ b/test/test_admin_cache_purge_race.sh
@@ -30,6 +30,8 @@ resolve_moqbin "$BINARY"
 source "$REPO/test/test_ports.sh"
 # shellcheck source=test_versions.sh
 source "$REPO/test/test_versions.sh"
+# shellcheck source=test_relay_lifecycle.sh
+source "$REPO/test/test_relay_lifecycle.sh"
 DATESERVER="$MOQBIN/moqdateserver"
 TEXTCLIENT="$MOQBIN/moqtextclient"
 
@@ -77,8 +79,11 @@ cleanup() {
     fi
     sleep 0.2
   done
-  wait "${PIDS[@]:-}" "${RELAY_PIDS[@]:-}" 2>/dev/null || true
+  reap_helpers "${PIDS[@]:-}"
+  local relay_failed=0
+  reap_relays "${RELAY_PIDS[@]:-}" || relay_failed=1
   rm -rf "$TMPDIR_SCRIPT"
+  (( relay_failed == 0 )) || exit 1
 }
 trap cleanup EXIT
 
--- a/test/test_admin_config.sh
+++ b/test/test_admin_config.sh
@@ -4,6 +4,8 @@ set -euo pipefail
 BINARY="${1:-$(dirname "$0")/../build/default/moqx}"
 # shellcheck source=test_ports.sh
 source "$(dirname "$0")/test_ports.sh"
+# shellcheck source=test_relay_lifecycle.sh
+source "$(dirname "$0")/test_relay_lifecycle.sh"
 LISTEN_PORT=$TEST_ADMIN_CONFIG_LISTEN
 ADMIN_PORT=$TEST_ADMIN_CONFIG_ADMIN
 ADMIN_URL="http://localhost:${ADMIN_PORT}/config"
@@ -16,8 +18,13 @@ fi
 TMPDIR=$(mktemp -d)
 MOQX_PID=""
 cleanup() {
-  [[ -n "$MOQX_PID" ]] && kill "$MOQX_PID" 2>/dev/null; wait "$MOQX_PID" 2>/dev/null || true
+  local relay_failed=0
+  if [[ -n "$MOQX_PID" ]]; then
+    kill "$MOQX_PID" 2>/dev/null || true
+    reap_relays "$MOQX_PID" || relay_failed=1
+  fi
   rm -rf "$TMPDIR"
+  (( relay_failed == 0 )) || exit 1
 }
 trap cleanup EXIT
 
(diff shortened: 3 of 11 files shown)


michalhosna (commented):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/590)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on afrind).

___
*[`test/test_relay_lifecycle.sh` line 10 at r1](https://reviewable.io/reviews/openmoq/moqx/590#-OzzcCwv3Ebr8qBzgS6r:-OzzcCwv3Ebr8qBzgS6s:benkl93) ([raw file](https://github.com/openmoq/moqx/blob/4f3c482513c5e4fa3ac98f3e0906eb5424ae33a4/test/test_relay_lifecycle.sh#L10)):*
> and SIGKILLs the survivors.



---

It feels to me that this helper should include the whole `for pid in pids; do kill.....` which is also duplicated. 
Not this just waits, feels a little bit like a footgun.

gmarzot:
*[`test/test_relay_lifecycle.sh` line 10 at r1](https://reviewable.io/reviews/openmoq/moqx/590#-OzzcCwv3Ebr8qBzgS6r:-P-ULVNuEm5aWmsMrKJD:bjc72ux) ([raw file](https://github.com/openmoq/moqx/blob/4f3c482513c5e4fa3ac98f3e0906eb5424ae33a4/test/test_relay_lifecycle.sh#L10)):*

> and SIGKILLs the survivors.



---

It feels to me that this helper should include the whole `for pid in pids; do kill.....` which is also duplicated. 
Not this just waits, feels a little bit like a footgun.
</blockquote></details>

from our friend

`reap_relays` is a bare `wait "$pid"` — unbounded in bash. The design leans on the relay's internal shutdown watchdog, but look at where that watchdog arms ([main.cpp:231-237](vscode-webview://0ub4klkrc2ncu5gjv155fim96rs7bhc6s4deu8m10q8kh96476sh/src/main.cpp#L231-L237)): the `_Exit(1)` thread is spawned **after `evb.loopForever()` returns** — i.e., only once shutdown has already begun. So:

-   Relay wedged **during teardown** → watchdog fires at 10s, exits 1, `wait` returns, test fails cleanly. ✅ Covered.
-   Relay wedged **before the loop exits** — a stuck event base that never processes the TERM, a deadlock mid-test, or a kill landing during startup before the handler is up → **watchdog never arms, `wait` blocks forever**. ❌ Not covered. The helper's comment ("a hung teardown (watchdog \_Exit(1))") slightly oversells the coverage.

And the backstops behind that hang are uneven, which is what makes it matter:

| Context | Backstop for a hung `wait` |
| --- | --- |
| `relay_chain` / `qmux_relay` | ctest TIMEOUT 120s / 60s — fine |
| all 8 `admin_*` tests | **no TIMEOUT set** → ctest default 1500s = 25 min each |
| `test_conformance.sh` in CI | **runs outside ctest entirely** (invoked directly by the conformance jobs) → only backstop is the job's `timeout-minutes: 180` |

That last row is the punchline: a pre-loop-exit wedge during a conformance run pins a hosted-runner slot for up to 3 hours — exactly this week's incident class, reintroduced through the test harness.

## Suggested fix (small, in `reap_relays` only)

Bounded reap with escalation: poll `kill -0` for ~30s (3× the relay watchdog's 10s, so the watchdog always gets first crack and clean TERM shutdowns are untouched); if the pid is still alive at the deadline, report `FAIL: relay did not exit within 30s (hung before teardown watchdog armed)`, SIGKILL it, then `wait` to collect. This is safe in bash — it reaps children as they die and holds their status for `wait`, so `kill -0` goes false the moment the relay exits, no zombie false-positives. The hang becomes a fast deterministic _failure_ (which is the PR's own goal) instead of a 25-minute-to-3-hour silent stall. Optionally, adding `TIMEOUT` properties to the 8 admin tests would be a good belt-and-suspenders in the same PR.

One smaller observation, take or leave: `reap_relays` calls `wait "$pid"` without `2>/dev/null`, so if a pid was already reaped (script bug, double-cleanup path) bash prints "not a child of this shell" and returns 127 → reported as relay failure. Arguably correct-by-loudness, but worth being aware it can mislabel a harness bug as a relay crash.

gmarzot (approved):
> @gmarzot reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/590)* status: all files reviewed, 1 unresolved discussion (waiting on michalhosna).

gmarzot:
*[`test/test_relay_lifecycle.sh` line 10 at r1](https://reviewable.io/reviews/openmoq/moqx/590#-OzzcCwv3Ebr8qBzgS6r:-P-UN5TY42RChTUiTQoQ:bcm7e52) ([raw file](https://github.com/openmoq/moqx/blob/4f3c482513c5e4fa3ac98f3e0906eb5424ae33a4/test/test_relay_lifecycle.sh#L10)):*
<details><summary><i>Previously, gmarzot (Giovanni Marzot) wrote…</i></summary><blockquote>

from our friend

`reap_relays` is a bare `wait "$pid"` — unbounded in bash. The design leans on the relay's internal shutdown watchdog, but look at where that watchdog arms ([main.cpp:231-237](vscode-webview://0ub4klkrc2ncu5gjv155fim96rs7bhc6s4deu8m10q8kh96476sh/src/main.cpp#L231-L237)): the `_Exit(1)` thread is spawned **after `evb.loopForever()` returns** — i.e., only once shutdown has already begun. So:

-   Relay wedged **during teardown** → watchdog fires at 10s, exits 1, `wait` returns, test fails cleanly. ✅ Covered.
-   Relay wedged **before the loop exits** — a stuck event base that never processes the TERM, a deadlock mid-test, or a kill landing during startup before the handler is up → **watchdog never arms, `wait` blocks forever**. ❌ Not covered. The helper's comment ("a hung teardown (watchdog \_Exit(1))") slightly oversells the coverage.

And the backstops behind that hang are uneven, which is what makes it matter:

| Context | Backstop for a hung `wait` |
| --- | --- |
| `relay_chain` / `qmux_relay` | ctest TIMEOUT 120s / 60s — fine |
| all 8 `admin_*` tests | **no TIMEOUT set** → ctest default 1500s = 25 min each |
| `test_conformance.sh` in CI | **runs outside ctest entirely** (invoked directly by the conformance jobs) → only backstop is the job's `timeout-minutes: 180` |

That last row is the punchline: a pre-loop-exit wedge during a conformance run pins a hosted-runner slot for up to 3 hours — exactly this week's incident class, reintroduced through the test harness.

## Suggested fix (small, in `reap_relays` only)

Bounded reap with escalation: poll `kill -0` for ~30s (3× the relay watchdog's 10s, so the watchdog always gets first crack and clean TERM shutdowns are untouched); if the pid is still alive at the deadline, report `FAIL: relay did not exit within 30s (hung before teardown watchdog armed)`, SIGKILL it, then `wait` to collect. This is safe in bash — it reaps children as they die and holds their status for `wait`, so `kill -0` goes false the moment the relay exits, no zombie false-positives. The hang becomes a fast deterministic _failure_ (which is the PR's own goal) instead of a 25-minute-to-3-hour silent stall. Optionally, adding `TIMEOUT` properties to the 8 admin tests would be a good belt-and-suspenders in the same PR.

One smaller observation, take or leave: `reap_relays` calls `wait "$pid"` without `2>/dev/null`, so if a pid was already reaped (script bug, double-cleanup path) bash prints "not a child of this shell" and returns 127 → reported as relay failure. Arguably correct-by-loudness, but worth being aware it can mislabel a harness bug as a relay crash.
</blockquote></details>

don't fear the reaper - more cowbell

afrind (commented):
> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/590)* status: all files reviewed, 1 unresolved discussion.

___
*[`test/test_relay_lifecycle.sh` line 10 at r1](https://reviewable.io/reviews/openmoq/moqx/590#-OzzcCwv3Ebr8qBzgS6r:-P-WKWSiA-d4C6Bivhuy:bfk4rka) ([raw file](https://github.com/openmoq/moqx/blob/4f3c482513c5e4fa3ac98f3e0906eb5424ae33a4/test/test_relay_lifecycle.sh#L10)):*
<details><summary><i>Previously, gmarzot (Giovanni Marzot) wrote…</i></summary><blockquote>

don't fear the reaper - more cowbell
</blockquote></details>

Sorry I had missed this discussion and merged right over it.  But we're in a better place now regardless. We can fix in a follow up if it's important.

PR #602, opened 2026-08-17
test: fix spurious SIGPIPE failure in cache-purge race check 5

gmarzot:
Main CI went red on macos after #600 merged: `admin_cache_purge_concurrency_test` reported `FAIL [5]: /metrics unresponsive after test` — but the relay was healthy.

**Root cause** — a latent shell bug in the check itself. Under `set -o pipefail`:

```bash
if ! echo "$METRICS" | grep -q "moqx_moqActiveSessions"; then
```

`grep -q` exits as soon as it matches. When `$METRICS` exceeds the pipe buffer, `echo` is still writing, takes SIGPIPE (exit 141), and `pipefail` propagates 141 as the pipeline status — so a *successful* match trips the FAIL branch. The job log shows the smoking gun immediately before the FAIL line:

```
test_admin_cache_purge_race.sh: line 279: echo: write error: Broken pipe
FAIL [5]: /metrics unresponsive after test (event loop wedged?)
```

The moxygen `c8cff3a` bump likely grew the /metrics body past the buffer threshold, exposing the race on the macOS runner (BSD grep buffering differs from GNU).

**Fix**: feed grep via herestring instead of a pipe — no pipe, no SIGPIPE, same semantics. The other `grep -q` pipes in the script consume small single-line bodies and are not affected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/602)

### DIFF
--- a/test/test_admin_cache_purge_race.sh
+++ b/test/test_admin_cache_purge_race.sh
@@ -276,7 +276,8 @@ echo "PASS [4]: Phase 2 purge returned valid response: evicted=$EVICTED"
 
 # 5. /metrics still readable (event loop not wedged).
 METRICS=$(curl -sf "$METRICS_URL" 2>/dev/null || echo "")
-if ! echo "$METRICS" | grep -q "moqx_moqActiveSessions"; then
+# herestring, not a pipe: grep -q exiting early would SIGPIPE echo under pipefail
+if ! grep -q "moqx_moqActiveSessions" <<< "$METRICS"; then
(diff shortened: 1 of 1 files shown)


afrind (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/602)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gmarzot).

PR #603, opened 2026-08-17
build: add --sync-moxygen-dir to move the moxygen checkout to the pin

afrind:
moxygen was a git submodule until recently, so git submodule update moved the local checkout to the recorded revision. It is a CPM pin now, and there is no equivalent command. A --moxygen-dir build compiles the checkout as it stands, so after a pin bump the checkout has to be moved by hand.

scripts/dev/sync-moxygen.sh moves a named checkout to MOXYGEN_REV, and configure.sh --sync-moxygen-dir runs it before configuring. It refuses to move a tree that is dirty, has an operation in flight, or sits on a commit no ref but HEAD names.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/603)

### DIFF
--- a/scripts/configure.sh
+++ b/scripts/configure.sh
@@ -3,7 +3,8 @@
 #
 # Usage:
 #   configure.sh [PROFILE] --moxygen (prebuilt-with-fallback|prebuilt|from-source)
-#                [--moxygen-dir DIR] [--clean] [-j N] [-DVAR=VALUE]...
+#                [--moxygen-dir DIR] [--sync-moxygen-dir] [--clean] [-j N]
+#                [-DVAR=VALUE]...
 #   PROFILE:                       default | san | tsan, or any preset from
 #                                  CMakeUserPresets.json; inherit `default` so
 #                                  binaryDir stays build/<name>. Default: default.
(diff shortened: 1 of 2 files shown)


michalhosna (commented):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/603)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on afrind).

___
*[`scripts/dev/sync-moxygen.sh` line 7 at r1](https://reviewable.io/reviews/openmoq/moqx/603#-P-KRzmfAAwSE1IlgHDT:-P-KRzmfAAwSE1IlgHDU:b-2e2bvx) ([raw file](https://github.com/openmoq/moqx/blob/1d5c9d176e24e0284e942893e6182c21831a2b82/scripts/dev/sync-moxygen.sh#L7)):*
> ```Shell
> #
> #   --strict    refuse loudly (non-zero exit) when the checkout cannot be moved.
> #               The default reports the reason and exits 0.
> ```

Nit: I feel like having --strict by default, but have something like `--ignore-errors` would be more friendly, I would call the current behaviour non-obvious. 

Even though I am not sure why not just use `sync-moxygen.sh || true`?

afrind (commented):
> @afrind reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/603)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on michalhosna).

___
*[`scripts/dev/sync-moxygen.sh` line 7 at r1](https://reviewable.io/reviews/openmoq/moqx/603#-P-KRzmfAAwSE1IlgHDT:-P-PbkiD2FuDEVvPg09J:bb7ohn6) ([raw file](https://github.com/openmoq/moqx/blob/1d5c9d176e24e0284e942893e6182c21831a2b82/scripts/dev/sync-moxygen.sh#L7)):*

Nit: I feel like having --strict by default, but have something like `--ignore-errors` would be more friendly, I would call the current behaviour non-obvious. 

Even though I am not sure why not just use `sync-moxygen.sh || true`?
</blockquote></details>

sgtm

afrind (commented):
> @afrind reviewed 2 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/603)* status: 0 of 2 files reviewed, all discussions resolved (waiting on michalhosna).

gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/603)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #612, opened 2026-08-20
build: derive the moxygen release tag from the pin

gmarzot:
## Why

The retention fix for the drift → from-source-fallback storms is openmoq/moxygen#371 (per-rev `snapshot-<sha12>` pre-releases, retained 30 days, instead of a rolling tag whose deletion orphans every older rev). This PR is the consumer side: resolve the pin **to that per-rev release directly**.

## What

`moqx_moxygen_release_tag()` no longer searches remote tags at all. The tag is a pure function of the pin — `snapshot-<first 12 of MOXYGEN_REV>` — and `FetchMoxygenPrebuilt` already verifies the resolved release's `target_commitish` against the pin before downloading, so a wrong or missing release fails loudly with guidance. Per review: no string-matching of another repo's tag names, no preference ordering, no `ls-remote` (and its git dependency is gone — `print-release-tag.cmake` is now a 0.1s no-network derivation, previously a network round-trip on every configure).

`MOXYGEN_RELEASE_TAG` keeps its role as the explicit override — release branches already pin it (docs/release.md), and it's the escape hatch for a rev whose per-rev snapshot aged out past the 30-day retention but that a durable `v*` release still carries. The 404 error message now explains all three causes (failed publish, pre-retention rev, aged-out snapshot) and both remedies.

## Testing

- Derivation unit-tested: mixed-case pin lowercases and truncates to `snapshot-<sha12>`; explicit `MOXYGEN_RELEASE_TAG` bypasses derivation.
- Live: `cmake -P cmake/print-release-tag.cmake` → `snapshot-1b32be12e4af` in 0.1s, no network.
- End-to-end fetch succeeds once #371 lands and a moxygen main push publishes the first per-rev release; until then the 404 path reports cleanly (equivalent to today's fallback behavior).

## Sequencing

Merge after (or with) openmoq/moxygen#371 — before #371's first published build, derivation points at a release that doesn't exist yet, so CI stays on the from-source fallback it's already on (e.g. right now: pin `1b32be1`, moxygen head `63062c2`, no tag on the pin).

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/612)

### DIFF
--- a/.github/workflows/version-release.yml
+++ b/.github/workflows/version-release.yml
@@ -130,6 +130,15 @@ jobs:
           fi
           echo "==> MOXYGEN_REV $MOX_PIN resolves to moxygen release $MOX_REF"
 
+          # Release branches must reference a permanent moxygen release:
+          # snapshots are pruned after 30 days, taking the release's moxygen
+          # reference and its reproducibility with them.
+          if [ "$BRANCH" != "main" ] && [[ "$MOX_REF" == snapshot-* ]]; then
+            echo "Error: $BRANCH resolves moxygen to the temporary snapshot $MOX_REF." >&2
+            echo "Set MOXYGEN_RELEASE_TAG in cmake/dependencies.cmake on $BRANCH to the moxygen v-tag for this release." >&2
+            exit 1
+          fi
+
           # Disagreement can only mean the operator meant to bump MOXYGEN_REV
           # first — the input cannot move the build off the pin.
           INPUT_REF="${{ inputs.moxygen_release }}"
--- a/cmake/FetchMoxygenPrebuilt.cmake
+++ b/cmake/FetchMoxygenPrebuilt.cmake
@@ -195,8 +195,11 @@ else()
     if(_http_code STREQUAL "404")
       message(FATAL_ERROR
         "moxygen: tag '${_rel_tag}' has no GitHub release, so it publishes no "
-        "prebuilt. Build from source:\n"
-        "  scripts/configure.sh --moxygen from-source && scripts/build.sh")
+        "prebuilt. The pinned rev's build failed, predates retained per-rev "
+        "snapshots, or its snapshot aged out (a v* release carrying the rev "
+        "still works: -DMOXYGEN_RELEASE_TAG=<tag>). Build from source:\n"
+        "  scripts/configure.sh --moxygen from-source && scripts/build.sh\n"
+        "or bump the pin in cmake/dependencies.cmake to a published rev.")
     elseif(_http_code STREQUAL "403" OR _http_code STREQUAL "429")
       message(FATAL_ERROR
         "moxygen: the GitHub API rate limit is exhausted for this IP (anonymous "
--- a/cmake/MoxygenRelease.cmake
+++ b/cmake/MoxygenRelease.cmake
@@ -59,8 +59,9 @@ function(moqx_moxygen_tags_at_pin OUT_VAR)
 endfunction()
 
 # Return the release tag whose assets belong to MOXYGEN_REV. GitHub keeps no
-# commit->release index, so this asks the remote which tags point at the pin, and
-# prefers an immutable v* release over a mutable snapshot-*.
+# commit->release index, so this asks the remote which tags point at the pin,
+# preferring immutable over mutable: a v* release, then the retained per-rev
+# snapshot-<sha12>, then a rolling snapshot-*latest.
 function(moqx_moxygen_release_tag OUT_VAR)
   # An explicit tag wins and needs no remote read; the fetch verifies the
   # release's commit against the pin either way.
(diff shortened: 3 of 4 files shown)


michalhosna (changes requested):
> Until now, the moment the fork drifted ahead of `MOXYGEN_REV`, the pinned rev lost its only tag and every CI job fell back to a from-source moxygen build — which is exactly what amplified yesterday's runner-queue mess.

But how is this PR this solving this? 

> @michalhosna reviewed 1 file and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/612#-:-P-VZNppEzM45VTmRzqy:bw2vw98)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on afrind and gmarzot).

___
*[`cmake/MoxygenRelease.cmake` line 64 at r1](https://reviewable.io/reviews/openmoq/moqx/612#-P-VZ_LG4gClImL0go5Y:-P-VZ_LG4gClImL0go5Z:baaekqd) ([raw file](https://github.com/openmoq/moqx/blob/7702c8bfead75ae4fa73c477d6f90250282ff6ff/cmake/MoxygenRelease.cmake#L64)):*
> ```CMake
> # commit->release index, so this asks the remote which tags point at the pin,
> # preferring immutable over mutable: a v* release, then the retained per-rev
> # snapshot-<sha12>, then a rolling snapshot-*latest.
> ```

I am not sure what is this bringing us? Shouldn't it don't matter which release is picked? If for given SHA there exists more releases with the correct artifact, surly it should not matter which one we pick? 

I don't love that we are now doing string matches on release names of and structurally independent repo so I would like to understand the benefits.

gmarzot:
Both points taken — adjusted in 9c81e5de.

**"How is this PR solving this?"** — it isn't, and the description overclaimed; fixed. The retention fix is openmoq/moxygen#371 (per-rev snapshots stop being deleted when main moves). This PR is the consumer-side complement: it makes the resolver's choice among matching tags *intentional* instead of an accident of ls-remote's ASCII ordering. A live illustration from an hour ago: main's pin is `1b32be1` but moxygen main moved this afternoon, so `snapshot-latest` now targets `63062c2` — the pin has **no** tag at all and every moqx CI job is back to from-source fallback. #371 is what fixes that; nothing in this PR could.

**"If several releases carry the artifact, why prefer one?"** — you're right that the assets are identical and the fetch verifies the commit regardless. The preference is about how long the resolved *name* stays valid, not content: the alias release is deleted and recreated on every moxygen push, so an alias resolved at configure time can dangle by fetch time (or on a retry), and release branches record the resolved tag via print-release-tag for reproducibility — a recorded alias goes stale on the next push, a v*/per-rev tag doesn't.

**String-matching an independent repo's tag names** — fair, and the code now matches much less structure: the `snapshot-[0-9a-f]{12}` pattern is gone. The single remaining rule is `-latest$` = mutable rolling alias, which is the publish contract our own fork defines in openmoq/moxygen#371 (openmoq/scripts/publish-artifacts.sh) — same org, both sides of the contract in our hands. v* preference over per-rev snapshots stays as before (pre-existing behavior).

gmarzot:
Took this a step further per discussion with @gmarzot, and it lands exactly where your comment points: **no tag matching at all**. Since #371 makes the per-rev release a guaranteed output of every moxygen build, the tag is a pure function of the pin — `snapshot-<sha12 of MOXYGEN_REV>` — derived, not searched. The fetch already verifies the release's commit against the pin before downloading, so correctness comes from the sha check, not from any name convention. ls-remote enumeration (and the git dependency, and a network round-trip per configure) deleted; `MOXYGEN_RELEASE_TAG` stays as the explicit override for v* releases / aged-out snapshots. Net −67 lines. See 48c8765b.

michalhosna (commented):
So now this PR rips out generic sha -> release lookup code and replaces it with leaning on moxygen release naming convention. 

Don't get me wrong, less code is generally good, but I am not sure it's a upgrade. What does it brings/solves besides ripping out code? 

> @michalhosna reviewed 1 file and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/612#-:-P-WAGX5A0M32Lvm0qf9:baiu09h)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on afrind and gmarzot).

___
*[`-- commits` line 40 at r3](https://reviewable.io/reviews/openmoq/moqx/612#-P-WBlwJ42L155nhQJ01:-P-WBlwJ42L155nhQJ02:bm5agyi):*
> no-network derivation

Why does that matter? You that then informating and go to fetch from github anyway, this doesn't work offline either way? 
___
*[`-- commits` line 39 at r3](https://reviewable.io/reviews/openmoq/moqx/612#-P-WBg1a6LKHDpKFyr51:-P-WBg1a6LKHDpKFyr52:b2sadcb):*
Is git dependency a problem?

gmarzot:
**We need** every build to reference the exact moxygen release its artifacts came from — and for version releases, that reference must be permanent.

**Because** the release workflow records the reference and trusts it: it decides whether to build portable tarballs from it, writes it into the release notes, and hands it to the portable builds. If the reference is wrong or temporary, a release quietly loses its portable tarballs today and its reproducibility in 30 days.

**Current** (`moqx_moxygen_release_tag` on main): it lists every tag pointing at the pin with `git ls-remote` and picks one of them. That was correct when every tag carried artifacts. After openmoq/moxygen#371, the `*-latest` tags are empty pointers — artifacts live only on `snapshot-<sha>` and `v*` releases — so the pick can land on an empty tag, and nothing fails loudly when it does.

**Proposed** (this PR): no picking. Daily builds construct the tag from the `MOXYGEN_REV` commit hash (`snapshot-<sha>`). Release builds name the moxygen `v*` tag explicitly via `MOXYGEN_RELEASE_TAG`, and as of 53a45c5f version-release fails with a clear message if a release-branch build didn't set it.

gmarzot:
Parking this. The discussion narrowed the real requirement down to the release path, and that's now covered by a much smaller change: #617 fails a release-branch dispatch whose moxygen reference resolves to a temporary snapshot, with no resolver changes. Daily CI is covered by retention (openmoq/moxygen#371) alone. What's left of this PR is simplification without urgency; the resolver edge it also addressed is documented in #618, and the branch (build/prefer-pinned-snapshot) keeps if that edge ever bites. Thanks @michalhosna for pushing on the justification — the narrower fix is the better outcome.

PR #617, opened 2026-08-20
version-release: require a permanent moxygen release on release branches

gmarzot:
Small, release-scoped guard extracted from the #612 discussion (which is being parked — see there).

Snapshot pre-releases are pruned after 30 days (openmoq/moxygen#371). If a release branch's `MOXYGEN_REV` resolves to a snapshot — i.e. nobody set `MOXYGEN_RELEASE_TAG` on the branch — the release builds fine today and loses its moxygen reference, and its reproducibility, a month later. Nothing fails loudly at any point.

This adds one check to `validate`: a non-main dispatch whose resolved moxygen ref is `snapshot-*` fails immediately with the fix in the message (set `MOXYGEN_RELEASE_TAG` in `cmake/dependencies.cmake` on the branch to the moxygen v-tag). Main-branch dispatches are unchanged: they build on the current snapshot and skip portable tarballs, exactly as today.

No resolver changes — this works with the tag resolution on main as-is.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/617)

### DIFF
--- a/.github/workflows/version-release.yml
+++ b/.github/workflows/version-release.yml
@@ -130,6 +130,15 @@ jobs:
           fi
           echo "==> MOXYGEN_REV $MOX_PIN resolves to moxygen release $MOX_REF"
 
+          # Release branches must reference a permanent moxygen release:
+          # snapshot pre-releases are pruned after 30 days, taking the
+          # release's moxygen reference and its reproducibility with them.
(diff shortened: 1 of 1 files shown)


gmarzot:
good grief ... am parking all this and letting things settle .. i will try the release workflow and see what happens

PR #621, opened 2026-08-22
admin: write JSON strings a run at a time

afrind:
JsonWriter::writeString appended one character at a time through the folly appender, even though track names and namespaces almost never contain a character that needs escaping. The walk runs on the relay executor, so that is the wrong default.

The scan now only breaks at characters that need escaping and pushes the unescaped run between them whole. Output is byte-identical -- checked differentially against the old loop over every single byte value and 200k random strings biased toward escapes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/621)

### DIFF
--- a/src/admin/JsonWriter.h
+++ b/src/admin/JsonWriter.h
@@ -120,26 +120,47 @@ private:
     app_.push(reinterpret_cast<const uint8_t*>(s.data()), s.size());
   }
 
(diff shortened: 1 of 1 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/621)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #622, opened 2026-08-22
admin: give route handlers an egress gate

afrind:
A route that streams a response in chunks has no way to stop producing while the transaction cannot write, so pausing only relocates the bytes into proxygen's egress queue instead of bounding them.

EgressGate makes proxygen's onEgressPaused/onEgressResumed awaitable on the admin EVB. AdminRequestHandler owns one, feeds it those callbacks, cancels it on requestComplete/onError, and passes it to the route as a fifth argument:

```
  if (!co_await egress->awaitDrainable()) {
    co_return; // request is going away, stop rather than send
  }
```

Every existing handler ignores the new parameter; no behavior change yet.  This will be used in subsequent commits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/622)

### DIFF
--- a/src/admin/AdminServer.cpp
+++ b/src/admin/AdminServer.cpp
@@ -10,6 +10,7 @@
 
 #include <folly/CancellationToken.h>
 #include <folly/io/IOBufQueue.h>
+#include <folly/io/async/EventBaseManager.h>
 #include <folly/logging/xlog.h>
 #include <proxygen/httpserver/RequestHandler.h>
 #include <proxygen/httpserver/RequestHandlerFactory.h>
@@ -17,6 +18,8 @@
 #include <proxygen/httpserver/ScopedHTTPServer.h>
 #include <proxygen/lib/http/HTTPMessage.h>
 #include <wangle/ssl/SSLContextConfig.h>
+
+#include "admin/EgressGate.h"
 
 namespace openmoq::moqx::admin {
 
@@ -28,7 +31,9 @@ namespace openmoq::moqx::admin {
 // ---------------------------------------------------------------------------
 class AdminRequestHandler : public proxygen::RequestHandler {
 public:
-  explicit AdminRequestHandler(RouteHandler handler) : handler_(std::move(handler)) {}
+  explicit AdminRequestHandler(RouteHandler handler)
+      : handler_(std::move(handler)),
+        egress_(std::make_shared<EgressGate>(folly::EventBaseManager::get()->getEventBase())) {}
 
   void onRequest(std::unique_ptr<proxygen::HTTPMessage> headers) noexcept override {
     req_ = std::move(headers);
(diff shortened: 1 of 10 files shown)


gmarzot (approved):
### 1. `awaitDrainable()` returns `true` after `shutdown()` — the guard is inverted exactly when it matters

```cpp
void shutdown() { done_ = true; paused_ = false; baton_.signal(cancelled); }

folly::coro::Task<bool> awaitDrainable() {
  while (paused_) { ... }   // paused_ was cleared by shutdown
  co_return true;           // "safe to send" — but the handler is deleted
}
```

`shutdown()` clears `paused_`, so any **subsequent** call takes the fast path and reports drainable. The doc comment says "False means the request is going away, so stop rather than send," and the PR body's idiom uses that return as the sole guard:

```cpp
if (!co_await egress->awaitDrainable()) { co_return; }
```

It only behaves correctly for a coroutine already parked at the moment of shutdown. A chunked producer that suspends elsewhere first — `/state` will, since `buildStateBody` hops to `workerEvb` — can come back to a `finish()`\-ed request, get `true`, and write to a `downstream_` whose `AdminRequestHandler` was `delete this`'d. One-line fix: `if (done_) { co_return false; }` above the loop.

### 2\. Thread affinity is documented but not enforced, and the trap is _destruction_

`~TimedBaton` does `XCHECK(!eventBase_ || eventBase_->isInEventBaseThread())` and, if still `notReady`, calls `signal()` which XCHECKs again. `eventBase_` is non-null here, so those are live.

The gate is handed to every route as a `shared_ptr` **by value**. If a handler passes it into work that runs on `workerEvb` and that frame drops the last reference, the destructor aborts on the wrong thread. `awaitDrainable()` from the wrong EVB likewise XCHECKs inside `wait()`. The existing handlers are already EVB-hoppy — worth either a `getEventBase()`\-based `XCHECK` at the top of `awaitDrainable()` (fails loudly at the await, not later at an unrelated frame destruction) or an explicit note in the header that the gate must not outlive the admin-EVB frame.

### 3\. "no behavior change yet" isn't quite true

`requestComplete()` previously did a bare `delete this`; it now routes through `finish()` and calls `requestCancellation()`. Probably benign and arguably more correct, but it's a real semantic change to the token contract documented in `AdminServer.h` ("signalled if the client disconnects") — the token now also fires on normal completion. Cheap to call out in the commit message so it's a decision rather than a side effect.

### 4\. Things I checked that are correct

-   The `finish()` ordering comment is right and load-bearing. `requestCancellation()` runs `TimedBaton`'s `CancellationCallback` **synchronously**, which posts the baton; cancelling before `shutdown()`/`delete this` means a waiter that resumes inline sees the token already set.
-   `reset()`\-on-pause doesn't strand a waiter: `folly::coro::Baton::reset()` clobbers `state_` and would drop a parked waiter, but `onPaused()` early-returns when `paused_` is already true, and a waiter can only be parked while `paused_` is true. Single EVB, no suspension point between the `while (paused_)` check and the park, so there's no window.
-   The `Status::notReady` case in the loop is real and handled: `signal()` schedules the resume, a later `reset()` sets `status_` back to `notReady`, and the waiter re-parks. The comment matches the code.
-   `timeout = 0ms` is sound. `wait()` guards timer scheduling on `timeout_.count() > 0`. And the rationale holds: proxygen's `HTTPTransaction` calls `refreshTimeout()` on write-sent, so a client reading _nothing_ trips the 60s `idleTimeout`, while a slow-but-progressing client keeps it alive — which is what you want.

### 5\. Nits

-   `paused()` is unused dead API.
-   `std::shared_ptr<EgressGate>` by value costs an atomic refcount per dispatch for a parameter all 7 handlers ignore; `const&` in the `RouteHandler` signature avoids it.
-   `make_shared<EgressGate>` on every request including `/info` and the 404 path.
-   Mixed styles in the diff: some handlers use `auto /*egress*/`, others spell `std::shared_ptr<EgressGate> /*egress*/`.
-   **No test.** `EgressGate.h` is header-only with non-obvious state; a gtest driving `onPaused`/`onResumed`/`shutdown` against a coroutine on a `ScopedEventBaseThread` is cheap and would have caught #1 directly. `test/` has no admin unit tests — only the `test_admin_*.sh` shell suites.

> @gmarzot reviewed 10 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/622#-:-P-f3Srg6B8jK1lYvSDA:b-ygluj6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

gmarzot:
image

afrind (commented):
* fixed shutdown bug
* added an XCHECK on awaitable thread
* update summary to reflect that there is a small behavior change
* fixed the shared_ptr copy cost (the alloc cost is peanuts) and aligned the styles
* decided not to add a test

> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/622#-:-P-fPeNb9EPHESZv0ri1:b-39vofg)* status: 0 of 10 files reviewed, all discussions resolved (waiting on gmarzot).

gmarzot (approved):
> @gmarzot reviewed 10 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/622)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #623, opened 2026-08-22
relay: hoist TrackStatsFilter's recent-group window into RecentGroupWindow

afrind:
The 3-group MRU window that lets countGroupIfNew tolerate interleaved subgroups was inlined in TrackStatsFilter as a raw array plus a count, so any other counter that needs the same tolerance would have to copy it.

It moves to relay/RecentGroupWindow.h as RecentGroupWindow, which owns the array, the LRU rotate and the size static_assert:

```
  if (recentGroups_.admit(groupID)) {
    ++counters.groups; // first sight of this group, or first since it aged out
  }
```

Same window size and same LRU behavior; TrackStatsFilter::kRecentGroups now aliases RecentGroupWindow::kSize.

Plan is to re-use in a future commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/623)

### DIFF
--- a/src/relay/RecentGroupWindow.h
+++ b/src/relay/RecentGroupWindow.h
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) OpenMOQ contributors.
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <algorithm>
+#include <array>
+#include <cstddef>
+#include <cstdint>
+
+namespace openmoq::moqx {
+
+// A most-recently-used window of group IDs, so counting distinct groups
+// tolerates the subgroups of several groups arriving interleaved. Bounds how
+// deep an interleave can be before a group is counted twice.
+//
+// LRU rather than insertion order: a group that keeps receiving subgroups stays
+// in the window however many other groups open alongside it.
+class RecentGroupWindow {
+public:
+  static constexpr size_t kSize = 3;
+
+  // True the first time a group is seen, or the first time since it aged out.
+  [[nodiscard]] bool admit(uint64_t groupID) {
+    for (size_t i = 0; i < count_; ++i) {
+      if (groups_[i] == groupID) {
+        std::rotate(groups_.begin(), groups_.begin() + i, groups_.begin() + i + 1);
+        return false;
+      }
+    }
+    count_ = std::min(count_ + 1, kSize);
+    std::rotate(groups_.begin(), groups_.end() - 1, groups_.end());
+    groups_[0] = groupID;
+    return true;
+  }
+
+private:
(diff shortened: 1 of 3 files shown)


gmarzot (approved):
`RecentGroupWindow` reads as a general-purpose container, but `kSize` is a hardcoded `static constexpr`, not a template parameter. If the second user wants a different window depth — and depth is the one thing a _different_ counter would plausibly want to tune — this becomes `template <size_t N>` later, churning both call sites and the `kRecentGroups` alias the tests depend on. Cheap to decide now:

```cpp
template <size_t N = 3>
class RecentGroupWindow { public: static constexpr size_t kSize = N; ... };
```

Since the whole justification is "plan is to re-use in a future commit," it's worth asking whether the future caller shares the depth. Given how the last two rounds went, decent odds that commit already exists in his Sapling stack and answers it.

## Smaller
-   **`[[nodiscard]]` on `admit()`.** It mutates on _both_ paths — LRU refresh on hit, insert on miss — so dropping the return silently loses a count while still updating the window. That's the exact misuse a bool-returning mutator invites, and the attribute is free.
    
-   **Name.** `admit` reads like a predicate but is unconditionally mutating; `touch()` or `insertIfAbsent()` would signal it. Related, if the depth question turns into a template: `RecentGroupWindow` bakes "group" into what is otherwise a generic tiny LRU set, so a second caller counting something other than groups would be reaching for an oddly-named type.
    
-   **Unit test.** Your usual shape for a `src/relay/` hoist is a unit test in the same commit — `InitialTrackState` and `LocalForwarderRegistry` both did. Is `RecentGroupWindow` different because `ActiveGroupSurvivesWindowChurn` and `RecountsGroupEvictedFromWindow` in `test/MoqxRelayTrackStatsTests.cpp` already pin the semantics, or is the test coming with the second caller?

> @gmarzot reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/623#-:-P-fAMum4NbuoEZe6gkV:b-ckb2sz)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

afrind (commented):
I don't think we need to make this a template or add dedicated unit test right now.  I like admit as a name.  Will add the [[nodiscard]]

> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/623#-:-P-fKD512qJNRbj35vWK:b-yglkub)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/623)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #624, opened 2026-08-22
admin: move the /state JSON visitors into their own header

afrind:
The two visitors that serialize /state were anonymous-namespace classes inside StateHandler.cpp, so nothing could construct one -- a test that wanted to check the JSON had to go through the HTTP handler.

Code motion only: they move to admin/JsonStateVisitors.h verbatim, and StateHandler includes it. /state output is unchanged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/624)

### DIFF
--- a/src/admin/JsonStateVisitors.h
+++ b/src/admin/JsonStateVisitors.h
@@ -1,0 +1,207 @@
+/*
+ * Copyright (c) OpenMOQ contributors.
+ * This source code is licensed under the Apache 2.0 license found in the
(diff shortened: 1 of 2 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/624)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #625, opened 2026-08-22
admin: serialize /state through a chunked writer

afrind:
The /state visitors held their own IOBufQueue and handed the finished body back at the end, so the whole document had to exist in memory before anything could be sent.

ChunkedJsonWriter owns the queue instead and hands a chunk to a sink once it crosses a threshold; the visitors call maybeFlush() at item boundaries:

```
  ChunkedJsonWriter writer([](auto chunk) { return send(std::move(chunk)); });
  JsonRelayContextVisitor visitor(writer);
```

Comma and nesting state stay in the JsonWriter, so a flush is invisible in the output and a chunk boundary can fall anywhere. StateHandler's sink still collects every chunk and sends the body whole at EOM, so /state is byte-identical -- verified against the previous commit, and again with the threshold forced to 1 so every item flushed separately.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/625)

### DIFF
--- a/src/admin/ChunkedJsonWriter.h
+++ b/src/admin/ChunkedJsonWriter.h
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) OpenMOQ contributors.
+ * This source code is licensed under the Apache 2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <folly/Function.h>
+#include <folly/io/Cursor.h>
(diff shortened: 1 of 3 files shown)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/625)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #627, opened 2026-08-23
relay: walk cached track stats in place

afrind:
/state's cache section went through MoqxCache::getTrackStats(), which copies every track name and allocates a group vector per track, so the walk could serialize the snapshot and then drop it.

forEachTrackStats hands the visitor a TrackStatsView borrowing the cache's own data, reusing one group vector across tracks; returning false stops the walk. The visitor's cache callback splits to match:

```
  visitor.onCacheBegin(bytes, now);
  cache_->forEachTrackStats([&](const auto& t) { return visitor.onCacheTrack(t); });
  visitor.onCacheEnd();
```

getTrackStats() stays for callers that want a snapshot, now implemented on top of the walk -- so MoqxCacheTest covers both. Tracks are visited in the same order, so the JSON is unchanged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/627)

### DIFF
--- a/src/MoqxCache.cpp
+++ b/src/MoqxCache.cpp
@@ -2412,23 +2412,30 @@ folly::Expected<folly::Unit, MoQPublishError> MoqxCache::cacheObjectAndUpdateByt
   return res;
 }
 
+void MoqxCache::forEachTrackStats(folly::FunctionRef<bool(const TrackStatsView&)> fn) const {
+  std::vector<GroupStats> groups; // reused across tracks; reserve only grows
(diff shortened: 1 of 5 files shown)


akash-a-n (approved):
> @akash-a-n reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/627)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #630, opened 2026-08-23
relay: count ingest on the relay executor, drop bogus subscriber counts

afrind:
/state read each subscription's counters off the forwarder, which in LocalForwarder mode belongs to another executor -- so the walk skipped it and reported zeros for every track. The two subscriber counts it reported were per-forwarder besides: "subscribers on this thread's forwarder", not on the relay.

TerminationFilter, already installed on both chain branches, becomes RelayIngestFilter and counts objects as they pass, into an IngestCounters the registry entry holds:

```
  void record(uint64_t group, uint64_t object); // objects, groups, largest
```

Distinct groups go through the same RecentGroupWindow TrackStatsFilter uses, so interleaved subgroups are not counted twice. /state reads those, so total_groups_received, total_objects_received and largest are right in every mode; subscribers and forwarding_subscribers are gone, and with them the json_exporter module that built per-track series out of /state -- moqx_track_subscribers from /metrics/track is the real number. Also drops getSubscribeWriteback, dead since the chain builder took over.

MoqxRelayStateTests walks /state in all three relay modes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/630)

### DIFF
--- a/docker/json-exporter/json-exporter.yml
+++ b/docker/json-exporter/json-exporter.yml
@@ -5,8 +5,8 @@
 # (see the moqx-state and moqx-info jobs in prometheus.yml).
 #
 # Per-track series come from the relay's own /metrics/track, not from here:
-# both emitted moqx_track_subscribers, and /state's entries are transient and
-# undercount under load.
+# /state no longer carries subscriber counts (they were per-thread forwarder
+# counts, meaningless in LocalForwarder mode).
 modules:
   moqx:
     metrics:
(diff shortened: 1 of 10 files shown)


akash-a-n (approved):
> @akash-a-n reviewed 10 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/630)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #632, opened 2026-08-23
relay: walk /state on the executor that owns each service

afrind:
MoqxRelayContext::dumpState ran the whole walk on the caller's thread, so /state read each relay's registry, namespace tree and cache from the admin EVB while the relay executor was writing them. It also hardcoded active_sessions to 0: a synchronous walk had no way to await the stats aggregate.

dumpState is now a coroutine. It copies the service list once, then hops to each service's getRelayExec() -- the single io thread for a service that has none -- for that service's walk, and to the upstream EVB to read the upstream state enum, which is a plain enum written on that thread. The visitor sees the same call sequence as before, just from a different thread per service, one at a time. active_sessions now comes from statsRegistry_->aggregateAsync().

ready() replaces the workerEvb() accessor for the handler's 503 check; admin has no other use for that EVB.

The response is still built whole and sent at EOM. test_admin_state.sh covers the route end to end.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/632)


akash-a-n (approved):
> @akash-a-n reviewed 8 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/632)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #633, opened 2026-08-23
admin: stream /state as the walk produces it

afrind:
/state serialized the whole document before sending a byte, so a relay with a large namespace tree or cache held the entire response in memory, and the client waited for the last service to be walked before seeing the first.

The walk now writes into an AsyncPipe that the admin EVB drains into a chunked response. Producer and consumer run concurrently: the producer migrates across relay executors, while the consumer stays on the admin EVB where sendBody is thread-safe and parks on the EgressGate whenever the transaction cannot write. A slow client therefore leaves the unsent bytes in the pipe rather than relocating them into proxygen's egress queue, where nothing would account for them. The walk itself never stalls: it is synchronous and never yields its executor, so the consumer cannot push back on it. StreamBudget bounds what that can cost instead, capping what may pile up in the pipe and failing the response past it. Headers wait for the first chunk, so a walk that throws before producing output is still reported as a 500 rather than an aborted 200.

Both visitors gain alive(), checked at section boundaries, so a client that hangs up stops the walk rather than formatting the rest for nobody.

StateStreamTest covers the visitors and chunk boundaries, StateResponseTest the consumer, budget and gate, and test_admin_state.sh the framing end to end.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/633)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -378,6 +378,7 @@ add_library(moqx_core STATIC
   src/admin/MetricsHandler.cpp
   src/admin/TrackMetricsHandler.cpp
   src/admin/StateHandler.cpp
+  src/admin/StateResponse.cpp
   src/stats/StatsRegistry.cpp
   src/stats/TrackStatsRegistry.cpp
   src/stats/MoQStatsCollector.cpp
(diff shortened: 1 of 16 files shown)


akash-a-n (approved):
> @akash-a-n reviewed 16 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/633)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

Issue #628, opened 2026-08-23
Build caches are over budget, so most CI jobs start cold


gmarzot:
Two problems, one fix.

## 1. The dependency cache is broken on hosted runners

`actions/cache` stores paths relative to the workspace. `~/.cache` is 3 levels
above a hosted workspace and 4 above the self-hosted one, but both derive the same
key `deps-ubuntu-22.04-x86_64`. An archive saved on linode replays as
`../../../../.cache` on a hosted runner, lands on `/home`, and fails to extract:

```
/usr/bin/tar: ../../../../.cache: Cannot mkdir: Permission denied
Failed to restore: "/usr/bin/tar" failed with exit code 2
```

**Hosted lanes re-clone the CPM sources on every run, today.** The ccache keys are
unaffected — they are lane-derived and each lane runs on one runner family.

## 2. We are 2.5x over the cache budget

24.7 GB against GitHub's 10 GB limit. The asan ccache alone is 14.4 GB, uploaded
from a self-hosted box that keeps `~/.cache` between jobs and gains nothing from
the round trip. Over budget, GitHub evicts least-recently-used, and `main`'s
caches go first — so PRs find nothing to restore.

## Expected benefit

Measured on real PR jobs, cache lines confirmed in each log:

| linux `ci pr` | cold | warm |
|---|---|---|
| prebuilt moxygen | 11 min | **2 min** |
| from-source fallback | 51 min | 11 min |

macOS 33 -> 7 and conformance 42-46 -> 7-8 in fallback mode. The asan lane is
4-11 min in every case — it uses the linode disk and never depends on the GitHub
cache, which is the argument for taking it out of the budget.

These were measured with the dependency cache already broken, so the real numbers
after a fix should be better.

## Fix

#636 — skip `actions/cache` on self-hosted (one condition fixes both problems),
bound ccache with an explicit max size, and hash compiler contents rather than
mtime so an apt/Homebrew compiler swap cannot serve stale objects.

moxygen has a different cache problem, filed as openmoq/moxygen#381.

gmarzot:
Dropping CCACHE_COMPILERCHECK=content in #658 — it misses against everything cached under the default scheme and the misses repeat, which took the linux Build step from 97s to 636s for the same 154 TUs.

afrind:
@gmarzot are we good here?

gmarzot:
Yes can close

PR #636, opened 2026-08-23
Cache only from hosted runners; bound and harden ccache

gmarzot:
Closes the correctness half of #628.

The self-hosted box keeps `~/.cache` between jobs, so uploading its caches to
GitHub buys nothing. Its ccache alone was 14.4 GB of the repo's 10 GB budget,
which evicts `main`'s entries and leaves every PR lane starting cold.

It also fixes a restore failure that is costing time today. `actions/cache`
stores paths relative to the workspace, and `~/.cache` sits 3 levels above a
hosted workspace but 4 above the self-hosted one:

| runner | workspace |
|---|---|
| hosted | `/home/runner/work/moqx/moqx` |
| linode | `/home/github-runner/actions-runner/_work/moqx/moqx` |

Both derive the same dependency-cache key, so an archive saved on linode replays
as `../../../../.cache` on a hosted runner, lands on `/home`, and fails with
`Cannot mkdir: Permission denied` for every entry. Hosted lanes have been
re-cloning the CPM sources on every run. The ccache keys are unaffected — they
are lane-derived and each lane runs on exactly one runner family.

Also sets `CCACHE_MAXSIZE` so the cache cannot grow back into the budget
(1500M hosted, 20G on the box with its own disk), and `CCACHE_COMPILERCHECK=content`,
since apt and Homebrew replace compilers under these runners and the default
`mtime` check would serve stale objects after a same-mtime swap.

This is items 1-4 of the target posture in #628. Items 5-6 (save-on-main-only,
staying under budget) follow separately; moxygen's side is openmoq/moxygen#381.

## Verifying

Cache behaviour cannot be exercised locally. On this PR's run, check that the
hosted lanes still show `Cache restored from key: ccache-Linux-...` and that the
`asan debug` job shows the two cache steps skipped. Afterwards the repo's cache
usage should drop from ~25 GB to roughly 7 GB.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/636)

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -28,7 +28,12 @@ runs:
 
     # The moqx CMakeLists routes the compiler through ccache automatically when
     # it is installed (install-system-deps.sh installs it).
+    #
+    # Hosted runners only: the self-hosted box keeps ~/.cache between jobs, so
+    # the round trip buys nothing and its ccache is large enough to crowd every
+    # other lane out of the repo-wide 10 GB cache budget.
     - name: Cache ccache
+      if: runner.environment == 'github-hosted'
       uses: actions/cache@v5
       with:
         path: ~/.cache/ccache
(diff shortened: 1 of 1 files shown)


afrind (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/636#-:-P0-l2k28W1_B7C2zZsL:bw0ykkx)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna).

PR #646, opened 2026-08-26
devops: fetch origin before sync-moxygen's containment check

afrind:
The containment check only reads local refs, so a moxygen checkout sitting on a commit that has since been pushed upstream is named by nothing locally and the detach is refused even though it would cost nothing -- the common shape right after moxygen work lands and the pin moves on with it. Fetch origin first, so the check sees the remote refs that do name it.

Best effort and non-fatal: it runs after the dirty/in-progress checks (which refuse without touching the network) and only on a goto that actually moves the checkout, since the HEAD == pin fast path returns earlier.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/646)

### DIFF
--- a/scripts/dev/sync-moxygen.sh
+++ b/scripts/dev/sync-moxygen.sh
@@ -76,6 +76,16 @@ for marker in rebase-merge rebase-apply MERGE_HEAD CHERRY_PICK_HEAD REVERT_HEAD 
   fi
 done
 
+# Both checks below read refs, and refs go stale: a checkout sitting on a commit that
+# has since been pushed upstream is named by origin/* on the server and by nothing
+# here, so the containment check refuses a detach that costs nothing. Placed after the
+# dirty and in-progress checks, which refuse without touching the network at all.
+#
+# Best effort, and with prompting off because the hook caller runs unattended: a fetch
+# that stopped to ask for credentials would hang every checkout.
+GIT_TERMINAL_PROMPT=0 git -C "$dir" fetch --quiet origin 2>/dev/null \
+  || echo "sync-moxygen.sh: could not fetch origin in $dir — falling back to local refs" >&2
+
 # What detaching can actually destroy is a commit reachable only from HEAD: nothing
 # else names it afterwards and the reflog is the only way back. Any ref containing
 # HEAD — a branch, a tag, a remote — keeps it, and `git checkout` returns to it, so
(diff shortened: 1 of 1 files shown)


gmarzot:
*[`scripts/dev/sync-moxygen.sh` line 90 at r1](https://reviewable.io/reviews/openmoq/moqx/646#-P0-qULy-7-zdS8J96O7:-P0-qULy-7-zdS8J96O8:b-whmij1) ([raw file](https://github.com/openmoq/moqx/blob/8a01ed5f9d96a1f51cb66572f97623f5482812bf/scripts/dev/sync-moxygen.sh#L90)):*
> ```Shell
> # Not fatal. Offline, the checks below just run against the refs already here and the
> # worst case is the conservative refusal they would have given anyway.
> git -C "$dir" fetch --quiet origin 2>/dev/null || true
> ```

The change is correct and worth landing. Two things I'd ask for:

**1\. The silent failure discards the diagnostic.** `2>/dev/null || true` means a fetch that fails for any reason — offline, dead remote, credential prompt — leaves no trace. The user then hits the containment refusal with no hint it was decided on stale refs, which is precisely the confusion this PR exists to remove. Keep it non-fatal, but emit one line to stderr on failure so the refusal below is interpretable.

**2\. Credential prompt / hang on the hook path.** The file header explicitly supports a post-checkout hook caller (`sync-moxygen.sh || true`). `git fetch` against a remote that wants credentials can block indefinitely, and a hook that hangs every `git checkout` is a bad failure. `GIT_TERMINAL_PROMPT=0` on that one command closes it cheaply. This is the one I'd treat as near-blocking, since it only bites people whose remote setup differs from the author's — the classic "works on my machine" shape.

Two smaller notes:

**3\. `--prune` cuts the other way, and it's a real decision.** Without it, a deleted remote branch leaves a stale `refs/remotes/origin/X` locally, which can make containment _pass_ for a commit that's no longer on the server — a false reassurance in the direction that loses work. The script's stated philosophy is refusing exactly what detaching would cost, which argues for accuracy in both directions. `--prune` makes it stricter and occasionally more annoying. Not obviously right, but worth being deliberate rather than defaulted.

**4\. The description slightly undersells the cost.** "not per-goto traffic" is true — the `HEAD == pin` fast path returns first. But among gotos that _do_ move, some previously needed no network at all, because the pin was already local from an earlier sync. Now every move fetches. Trivial for a repo this size; just not quite "no new traffic".

Optionally: ten lines of comment for one line of code, and much of it is "not before / Not fatal / not per-goto". Since the house standard is saying what the code does rather than what it isn't, it could compress to about three. Low priority and consistent with the file's existing voice, so take it or leave it.

gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/646)* status: all files reviewed, 1 unresolved discussion (waiting on afrind).

gmarzot (approved):
> @gmarzot reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/646)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #648, opened 2026-08-27
build: root XLOG categories and debug paths past ccache

michalhosna:
Sharing one ccache across work trees was quietly breaking correctness.

  - The same TU logged under moqx.ServiceMatcher in a non-ccache build and under a src-rooted category in a ccache build. Filtering logs by category could not be relied on.
  - A cached object carried the debug paths of whichever worktree compiled it first. Debugging a binary in one worktree could open a different worktree's source.

Cause: ccache with base_dir rewrites the source path it forwards to the compiler but leaves -f*-prefix-map untouched, so the map stopped matching. The block comment carries the detail.

Fixes: An object now embeds no worktree path. The shared cache becomes sound

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/648)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,14 +286,28 @@ CPMAddPackage(
         OPTIONS "REFLECTCPP_YAML ON" "REFLECTCPP_BUILD_TESTS OFF"
 )
 
-# --- XLOG category rooting ---
+# --- XLOG category and debug path rooting ---
 # Sources live under src/, but folly XLOG categories should read moqx.MoqxRelay
-# rather than src.MoqxRelay. FOLLY_XLOG_STRIP_PREFIXES covers what the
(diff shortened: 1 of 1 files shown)


gmarzot:
Should the build dir get a relative map too, or do generated headers not go through base_dir?

michalhosna:
They do go through base_dir. 

ccache makes paths relative to the build dir, so
- a generated header comes out as generated/Foo.h, no prefix left. 
- src/ sits outside the build dir, so it comes out as  ../../src/Foo.cpp and still has a prefix to map.

gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/648)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna).

PR #654, opened 2026-08-28
relay: defer CrossExecFilter destruction to its target executor

afrind:
CrossExecFilter's methods enqueue [this]-capturing lambdas on targetExec_, but nothing keeps the filter alive until they run — owners release it from whatever executor they happen to be on. It can be freed with calls still queued, and the lambda then reads downstream_ off freed memory, seen as a SIGSEGV in beginSubgroup on a v0.3.1 relay.

The sibling filters anchor themselves with selfGuard_, but that needs a terminal call and a TrackConsumer has none that is guaranteed. Destruction is guaranteed, so create() installs a deleter that posts the delete to targetExec_ behind every lambda already queued, and PrivateTag makes create() the only way to build one.

```
  auto filter = CrossExecFilter::create(targetExec, downstream);
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/654)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -658,7 +658,7 @@ Subscriber::PublishResult MoqxRelay::publishFromPublisherExec(
   // crossExecFilter is a channel subscriber for the relay exec
   // registerPublishOnRelayExec completes wiring the chain (topNFilter → terminationFilter →
   // cache).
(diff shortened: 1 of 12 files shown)


gmarzot (approved):
> @gmarzot reviewed 12 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/654)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

Issue #394, opened 2026-06-09
Simplify Top-N complexity with N+X Filtering


suhasHere:
## Discussion: Simple N+X Top-N Filter Design for TRACK_FILTER

**Summary:** Proposing the Simple N+X ranking algorithm as the production path for MOQ TRACK_FILTER evaluation, replacing the current waterline-key approach.

---

### What we have today: Waterline-Key Approach

The current `PropertyRanking` implementation maintains a per-subscriber **waterline position** — for each subscriber requesting top-N, we pre-compute the RankKey of the Nth non-self track in the sorted order. A track is forwarded if its rank key is >= the subscriber's waterline.

**At a vey highlevel the design works as below:**

1. All tracks are sorted by (value DESC, arrival sequence ASC)
2. On every property value update, `computeWaterlineKey()` iterates the map for each affected subscriber to find the Nth non-self entry
3. The waterline key is stored per (subscriber, maxSelected) pair
4. Selection is O(1) at egress time: compare track's key against stored waterline

**Existing optimizations:**

- **Value deduplication**: If the incoming value equals the current value, the update is a no-op — no recomputation triggered
- **Threshold fast-path (`crossesThreshold`)**: If the updated value doesn't cross any subscriber's boundary (stays above or below all waterlines), recomputation is skipped entirely

These two optimizations eliminate work in the common case where values don't change or don't affect ranking boundaries. The challenges below apply specifically to what happens **when the threshold IS crossed**

---

### Challenges with the Waterline Approach at Scale

| Challenge | Impact |
| --- | --- |
| **O(S × T) update cost when threshold crossed** | When a value change crosses a waterline boundary, recomputation scans the sorted map once per subscriber to find their Nth non-self entry. With 90 panelists and 1,200 subscribers, a single boundary-crossing value change triggers up to 1,290 waterline recomputations. |
| **Per-subscriber state that cannot be shared** | Each subscriber maintains its own waterline key + selected-tracks set. Because each panelist has a different "self" to exclude, subscribers with the same N *still* have different waterlines — the state is inherently per-subscriber with no sharing possible. |
| **Mutex contention on the hot path** | Waterline recomputation and egress reads share state under a mutex. The mutex is held for the *entire duration* of waterline recomputation (scanning S subscribers). At high update rates with frequent boundary crossings, this blocks the forwarding path. |
[...]

PR #662, opened 2026-09-01
cache: port upstream's fetch range validation into MoqxCache

gmarzot:
Draft, for discussion. This is our reading of Alan's note: *"We have forked the cache, we can revert all OpenMOQ changes there. But that change needs to make it to moqxcache or the new conformance tests will fail."*

This is the moqxcache half. The companion PR is openmoq/moxygen#396, which reverts moxygen's copy of the cache to upstream.

## Why

moxygen #394 reworked upstream's `MoQCache` and shipped new conformance tests with it. The relay serves fetches from `MoqxCache`, our fork, which didn't get the change.

moqx PR #661 bumps the moxygen pin across that commit, and all three conformance jobs fail — they ran 3h before being cancelled. The previous pin bump, #660, passed all three in about 7 minutes each. Same jobs, only the pin differs.

From the #661 log:

```
[Test 61] FETCH range B-C, C cached
  Request: fetch
  FETCH {1,0}..{1,0}
  ... (no further output, cancelled after 3h)
```

`{1,0}` on the wire asks for all of group 1. The cached path already folded that into the exclusive `{2,0}`; the miss path did not, so its range iterator was constructed with `{1,0}` as both start and end — an empty range that never completes. Test 42 also crashes with exit 139.

## What this ports

| From upstream #394 | Effect |
|---|---|
| `toExclusiveEnd()` | folds the wire form once, used by the miss path and the cached path alike; `fetchUpstream` puts the wire form back for the outgoing request |
| `checkInRange()` | rejects anything upstream sends outside the range it was given, on `object`, `beginObject`, `endOfGroup`, `endOfTrackAndGroup` |
| `beginObjectAccepted_` | a payload following a refused `beginObject` is dropped instead of attaching to whatever `group.objects[...]` default-constructs |
| `FetchWriteback` range source | takes `minLocation`/`maxLocation` from the iterator instead of separate `start`/`end` arguments — the old call passed the *wire* form into `fetchesInProgress` |
| idempotent `eraseAndMakeGroupEvictable()` | the destructor can call it unconditionally |

Kept moqx-local: `shouldSkipCaching()` in place of upstream's `evicted` check, and the formatting.

## Tests

Ported upstream's three new tests and their range fix:

- `TestFetchMissWholeEndGroup` — the miss path agrees with the cached path on `{N,0}`
- `TestFetchMissUpstreamOverrunsRange` — an overrun is rejected, not cached or forwarded
- `FetchWritebackObjectPayloadAfterRefusedBeginObject` — the payload guard
- `TestUpstreamFetchUsingBeginObjectAndObjectPayload` — range widened to `{0,2}`, since it serves two objects

**One judgment call worth a look.** `FetchWritebackUpdateInProgressBoundaryCollision` is moqx-local, not upstream's. It pairs two fetches meant to abut at group 10, written when the miss path left `{10,0}` un-normalized. With the fix, A's wire end of `{10,0}` means *all of* group 10, so A and B overlap instead of abutting and the test hangs. I changed A's wire end to `{9,0}`, which is the exclusive `{10,0}` B starts on, preserving the adjacency the test was written for. If the original numbers were deliberate, this needs a different fix.

## Validation

`moqx_cache_test`: 156/156 pass locally, including the four above.

Conformance is **not** validated here. This branch is on the current pin (`37fed63`), which predates the new tests, so CI will run the old conformance suite and prove nothing. Per our convention pins move only through sync PRs, so this doesn't bump it. To actually validate, either land this and re-run #661, or move this commit onto `sync-moxygen/71ed936`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/662)

### DIFF
--- a/src/MoqxCache.cpp
+++ b/src/MoqxCache.cpp
@@ -21,6 +21,21 @@ using namespace moxygen;
 // Cap on prior-gap validation scans to avoid O(n) work for huge gaps.
 constexpr uint64_t kMaxGapValidation = 100;
 
+// On the wire an end object of 0 asks for all of the end group; every other
+// value is already one past the last object.  Everything inside the cache
+// works in the plain exclusive form this returns.
+AbsoluteLocation toExclusiveEnd(AbsoluteLocation wireEnd) {
+  if (wireEnd.object != 0) {
+    return wireEnd;
+  }
+  auto nextGroup = wireEnd.nextGroup();
+  if (!nextGroup) {
+    XLOG(ERR) << "Fetch end group=" << wireEnd.group
+              << " at max, using kLocationMax as exclusive end";
+  }
+  return nextGroup.value_or(kLocationMax);
+}
+
 // Compute the next iterator position after a gap when iterating descending
 // (NewestFirst). Groups iterate high-to-low, but objects within a group go
 // low-to-high. Returns std::nullopt if the gap extends past the iteration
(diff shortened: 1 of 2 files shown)


afrind (approved):
One thing the PR missed

  FetchWritebackUpdateInProgressDoubleCall (test/MoqxCacheTest.cpp:1521) silently stops testing its regression. Same class of problem as
  the collision test, but it wasn't adjusted.

  It fetches getFetch({0,0}, {2,0}) on a new track. Post-fix maxLocation becomes {3,0} instead of {2,0}. endOfGroup(1,0,0,fin=true)
  still leaves the iterator at {2,0}, but now current < maxLocation is true, so updateInProgress (MoqxCache.cpp:973) takes the re-key
  branch instead of the erase branch. fetchInProgressIt_ is therefore never end() when endOfFetch() fires the second updateInProgress,
  so the unguarded dereference at line 968 — the entire point of the test — is never reached. The test still passes; it just proves
  nothing, and its comment now describes a scenario that can't happen.

  Fix is the same shape as the one already applied: wire end {1,0} (exclusive {2,0}) in both the expectUpstreamFetch at :1539 and the
  getFetch at :1549. That restores maxLocation == {2,0}, endOfGroup(1,0,0) stays in range, the iterator lands exactly on maxLocation,
  branch 3 erases, and the second call hits end() as intended.

but I'm taking now and we'll fixup later

> @afrind reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/662#-:-P0XdArY81xpzNkik0bg:b-p7beia)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gmarzot).

PR #665, opened 2026-09-02
BugReport: cover fetch range accounting in the cache

michalhosna:
Two tests in their own target, one per rule:

- a fetch stream that FINs with no object bounds its non-existence inference at the Largest Location in FETCH_OK; the range past that Largest resolves against upstream;
- a whole-group fetch (End Location object 0, §10.12.1) on a track the cache does not hold keeps every object it delivered.

Problems:
  - The cold path in MoqxCache::fetch passes the spec-encoded end {0,0} straight into FetchWriteback and FetchRangeIterator. The warm path normalizes it to an exclusive {1,0} first. The iterator ends where it starts, so each later object's gap marking wipes the previous one. All three objects are cached, only the last one is served.
  - FetchWriteback::endOfFetch marks gaps up to the iterator's end and never consults the upstream FETCH_OK End Location. The TODO in fetchUpstream already names this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/665)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -414,3 +414,14 @@ moqx_add_gtest(moqx_cache_test
     Folly::folly_coro_gtest_helpers
     Folly::folly_io_async_scoped_event_base_thread
 )
+
+moqx_add_gtest(moqx_cache_fetch_range_test
+  SRCS MoqxCacheFetchRangeTest.cpp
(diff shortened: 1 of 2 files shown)


afrind:
Part of this was already fixed in main.  The other half fix is in #755

PR #666, opened 2026-09-02
BugRepro: upstream fetch-credit test @draft18

michalhosna:
Problem: Against current moxygen the test stalls at ~14 of 22: moxygen leaves the publisher half of a completed FETCH's request stream open (`fetchOk` keeps it open, `fetchComplete` never FINs it), the stream never reaches both-directions-terminal and its MAX_STREAMS credit is never returned.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/666)


afrind:
There's a fix in https://github.com/openmoq/moxygen/pull/461, which will also get upstream, eventually

PR #668, opened 2026-09-02
ci: require a published snapshot before bumping the moxygen pin

gmarzot:
The sync pins the newest moxygen main commit whether or not its snapshot exists
yet. moxygen `ci-main` takes ~80 minutes to build and publish, so a sync landing
inside that window pins a rev nothing can consume, and every moqx lane compiles
the Meta stack from source instead.

## Observed

2026-08-28, moxygen `68465b9`:

```
16:15  moxygen ci-main starts building 68465b9
16:21  moqx sync PR opens, pinned to 68465b9      <- 69 min early
17:30  moxygen publishes snapshot-68465b9ffa19
```

Every lane in that PR fell back to source: linux 49 min, conformance 45/45/37,
macOS 39, asan 34 — roughly four hours of runner time for a one-line pin bump.
Build-step times in the conformance lanes were 2062s, 2590s and 2620s against
286s, 212s and 280s on a normal pin bump.

## Change

Scheduled and manual runs walk back from main until they find a rev whose
`snapshot-<sha12>` release has assets, rather than taking the newest rev
unconditionally.

- **Nothing found in 20 commits** — skip with a warning and no PR. A publish
  backlog is transient; a red sync every hour is not useful.
- **`repository_dispatch`** keeps its payload rev but errors if the assets are
  missing. The publish is what raises that event, so a miss means it reported a
  success it did not deliver.

The cron cascade already tries to sequence this by spacing moxygen's sync at 04:23
against moqx's at 08:23. That only holds when schedules fire on time, and they
have been arriving hours late; the case above was a manual dispatch, which lands
wherever it lands. This makes the ordering a property of the data rather than of
the clock.

## Verifying

`gh api` prints its 404 body on stdout, so the predicate takes the exit code as
the answer and only compares a value that is actually a count — the obvious
`|| echo 0` form appends to the error JSON and fails the comparison with
`integer expression expected`, once per skipped commit.

Checked against live data: the predicate returns true for the current main rev
(14 assets) and false, silently, for a nonexistent one. Selection is a no-op today
because the newest rev already has assets, which is the intended behaviour — it
only diverges inside a publish window.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/668)

### DIFF
--- a/.github/workflows/moxygen-sync.yml
+++ b/.github/workflows/moxygen-sync.yml
@@ -122,11 +122,36 @@ jobs:
           CURRENT_SHA=$(cmake -DPIN=MOXYGEN_REV -P cmake/print-pin.cmake)
           echo "current_sha=${CURRENT_SHA}" >> "$GITHUB_OUTPUT"
 
+          # A rev is consumable once ci-main publishes its snapshot. gh prints
+          # its 404 body on stdout, so the exit code is the answer.
+          has_assets() {
+            local n
+            n=$(gh api "repos/openmoq/moxygen/releases/tags/snapshot-${1:0:12}" \
+                  --jq '.assets | length' 2>/dev/null) || return 1
+            [ "${n:-0}" -gt 0 ]
+          }
+
           if [ "${{ github.event_name }}" = "repository_dispatch" ]; then
             TARGET_SHA="$CLIENT_PAYLOAD_SHA"
+            # The publish raises this event, so a miss means it reported a
+            # success it did not deliver.
+            if [ -n "$TARGET_SHA" ] && ! has_assets "$TARGET_SHA"; then
+              echo "::error::moxygen ${TARGET_SHA:0:12} dispatched an update but publishes no snapshot assets"
+              exit 1
+            fi
           else
-            # Manual/scheduled trigger: latest openmoq/moxygen main
-            TARGET_SHA=$(gh api repos/openmoq/moxygen/commits/main --jq '.sha')
+            # Newest openmoq/moxygen main whose snapshot has published.
+            TARGET_SHA=""
+            for sha in $(gh api "repos/openmoq/moxygen/commits?sha=main&per_page=20" --jq '.[].sha'); do
+              has_assets "$sha" && { TARGET_SHA="$sha"; break; }
+              echo "skipping ${sha:0:12} — snapshot not published yet"
+            done
+            # Nothing publishable: hand the current rev to the up-to-date path
+            # below rather than failing, since a publish backlog is transient.
+            if [ -z "$TARGET_SHA" ]; then
+              echo "::warning::No moxygen main commit in the last 20 has published snapshot assets"
+              TARGET_SHA="$CURRENT_SHA"
+            fi
           fi
 
           # Guard: TARGET_SHA is spliced into cmake/dependencies.cmake by sed


afrind (commented):
Meat Proxy:

  Four problems.
  
  ## Manual runs walk back, and they shouldn't

  `workflow_dispatch` falls into the same branch as the cron, so it walks back too.
  My workflow is: land a moxygen change, then start a sync. That always lands inside
  the 80 minute publish window, so the walk-back will always skip the commit I just
  landed and pin the one before it.

  It gets worse from there, because PHASE A allows one sync PR at a time:

  - I land moxygen N and run sync.
  - N has no snapshot yet, so the PR opens for N-1.
  - That PR now blocks every later sync.
  - I can't re-run to get N until the N-1 PR merges.
  - My change shows up on the next cron, a day later.

  Today the first try pins N. So for a manual run this is a step backwards.
  
  I'd rather the PR get created against the newest commit and have CI wait for the
  snapshot than have sync quietly pick an older commit.

  ## The repository_dispatch error is based on something that isn't true anymore

  The description says the publish is what raises the event, so a missing snapshot
  means the publish reported a success it didn't deliver. moxygen doesn't raise it
  anymore. `omoq-ci-main.yml:412` says the sync is cron-driven "rather than per-merge
  repository_dispatch from here", and grepping moxygen for `repository_dispatch`
  turns up nothing but that comment.

  So the only thing that raises `moxygen-update` is a person pushing a change through
  by hand. That's the urgent case, and this change makes it fail with no PR.
  
  ## `.assets | length > 0` is true while assets are still uploading

  `publish-artifacts.sh` creates the release first (line 209), then uploads assets one
  at a time in a retry loop (line 219). There's a window where the release has 1 of 14
  assets and this check calls it published. We could pin a rev whose arm64 tarball
  hasn't landed and watch a lane fail to fetch it.

  Same window on a ci-main re-run, which deletes and recreates the release (line 195).
  An already-published rev reads as unpublished while that happens.

  ## A gh failure looks the same as "not published"
  
  `|| return 1` catches 404, rate limits, 5xx and network errors all the same, and the
  caller just moves on to the next commit. A blip on the newest rev silently pins an
  older one, and nothing in the log says why.

  Small one: on a manual run the "nothing publishable" fallback gives you a green
  check, no PR, and a warning nobody reads. The Slack and email notifiers only cover
  the blocked case. If manual keeps any skip behavior it should be loud.

  ## Options for the manual case

  1. **Split the trigger three ways.** `schedule` keeps the walk-back. `workflow_dispatch`
     and `repository_dispatch` take the tip unconditionally. Smallest change, restores
     today's behavior on the two paths a human drives, does nothing about CI cost.

  2. **Open the PR as a draft when the snapshot isn't there, promote it when it is.**
     `ci-pr` already skips conformance on drafts and already re-triggers on
     `ready_for_review`. But `build` isn't draft-gated, so as it stands this only saves
     the three conformance lanes, about 127 minutes of the 250. It'd need the same gate
     on `build`. Promotion can be a small scheduled job that finds the draft sync PR and
     checks the tag.
  
  3. **Have the sync job wait.** Take the tip, poll for the snapshot, then create the PR.
     One runner idle for up to 80 minutes instead of 250 minutes spread across lanes.
     Simple, but no PR exists while it waits.

  4. **Add `workflow_dispatch` inputs**, e.g. `rev` (empty = tip) and a `require_snapshot`
     boolean. Lets the cron and the human ask for different things without hardcoding
     the difference into the event name.

> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/668#-:-P0cQr_S0INe_cYMg8xz:b-lkpnja)* status: 0 of 1 files reviewed, all discussions resolved (waiting on michalhosna).

gmarzot:
superseded - closing

PR #677, opened 2026-09-04
relay: stop releasing an ending session on the relay executor

afrind:
onSessionEnd moved the shared_ptr<MoQSession> into the lambda it posted to relayExec_, so when that was the last reference the session was destroyed on the relay executor instead of its own. The lambda now captures the raw pointer it uses as the map key plus a weak_ptr, neither of which keeps the session alive. It erases the hop ID entry only when the stored weak_ptr and the ending session share an owner, so a new session that lands on a recycled address keeps its own entry.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/677)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/677)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #679, opened 2026-09-04
cache: record the upstream end of track

afrind:
A FETCH response has no END_OF_TRACK object, so the flag on the upstream FETCH_OK is the only place a fetch learns the track ends inside the range it asked for. The cache read that reply for its extensions and dropped the rest. Replaying the same track from cache then told the client End Of Track 0 and sent another pointless request upstream for the range past the last object.

The cache now folds the upstream end of track back into the track it is filling, on both the cold miss path and the tail fetch path. CacheTrack::updateLargest() also picks up the flag when the end of track is an object it already has.

Ported from moxygen commit 1c75b1c6, including its test coverage.

Claude-Session: https://claude.ai/code/session_01MRY5gbyL3XEngVhotspy6h

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/679)

### DIFF
--- a/src/MoqxCache.cpp
+++ b/src/MoqxCache.cpp
@@ -433,12 +433,15 @@ MoqxCache::CacheTrack::updateLargest(AbsoluteLocation current, bool eot) {
       endOfTrack = eot;
     }
     largestGroupAndObject = current;
-  } else if (eot && current != *largestGroupAndObject) {
-    // End of track is not the largest
-    XLOG(ERR) << "Malformed track, eot is not the largest object";
-    return folly::makeUnexpected(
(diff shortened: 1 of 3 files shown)


mondain (approved):
:lgtm:

> @mondain reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/679#-:-P0icPw9A6noVtjVbX2f:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #681, opened 2026-09-04
cache: forward the datagram lastInGroup flag to the downstream consumer

afrind:
SubscribeWriteback::datagram() took lastInGroup but dropped it on both the skip-caching and the caching path, so the END_OF_GROUP marker never reached the downstream consumer for datagram objects.

Ported from moxygen commit 1074cb5b (the MoqxCache-equivalent hunk only — the MoQFramer/MoQTypes/MoQSession portions of that commit are already picked up via the vendored moxygen library, which is pinned past that commit). No upstream cache test exists for this one (its coverage lived in MoQFramerTest.cpp, not applicable here), so this adds regression coverage for both the caching and skip-caching paths.

Claude-Session: https://claude.ai/code/session_01MRY5gbyL3XEngVhotspy6h

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/681)

### DIFF
--- a/src/MoqxCache.cpp
+++ b/src/MoqxCache.cpp
@@ -895,11 +895,10 @@ public:
   }
 
   folly::Expected<folly::Unit, MoQPublishError>
-  datagram(const ObjectHeader& header, Payload payload, bool /*lastInGroup*/ = false) override {
+  datagram(const ObjectHeader& header, Payload payload, bool lastInGroup = false) override {
     if (track_.shouldSkipCaching()) {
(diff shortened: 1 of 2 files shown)


mondain (approved):
> @mondain reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/681)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #683, opened 2026-09-04
relay: hop peer subscription calls to the forwarder's executor

afrind:
In RelayExec mode relayExec_ owns the forwarder, but startPublish handed the session a MoQForwarder::Subscriber unwrapped, so unsubscribe() and requestUpdate() reached forwarder state and registry_ on the session's io thread while the relay thread was touching the same track. Concurrent erase corrupts the F14 table. The handle is now wrapped so both calls hop first, the way the subscribe direction already wraps the handles it returns:

  peerHandle = std::make_shared<CrossExecSubscriptionHandle>(subscriber, relayExec_);

The wrapper copies inner's SubscribeOk because hasSubscribeOk() is non-virtual, and MoQSession seeds a track's largest from it. TrackPublisherImpl::unsubscribe now resets its subgroups before the forwarder drops the subscriber; the forwarder reaches the same streams through CrossExecFilter, which already deferred them, so the second reset was redundant either way. The fixture gains peerUnsubscribe(), which delivers a peer UNSUBSCRIBE the way MoQSession::onUnsubscribe does.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/683)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -11,6 +11,7 @@
 #include "relay/ChannelSubscriber.h"
 #include "relay/CrossExecFilter.h"
 #include "relay/CrossExecForwarderCallback.h"
+#include "relay/CrossExecSubscriptionHandle.h"
 #include "relay/InitialTrackState.h"
 #include "relay/LocalForwarderCallback.h"
 #include "relay/NullConsumers.h"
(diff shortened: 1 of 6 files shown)


mondain (approved):
:lgtm:

> @mondain reviewed 6 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/683#-:-P0i2N5N6xamE68yOG_4:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #684, opened 2026-09-04
fix: route SUBSCRIBE/FETCH past publisher-less namespace nodes

afrind:
Port of moxygen a682b57a ("Route SUBSCRIBE past publisher-less namespace nodes"). findPublisherSession() walked the namespace tree and read the publisher off whichever node the walk stopped at. PUBLISH and SUBSCRIBE_NAMESPACE also create nodes, and those carry no publisher, so a lookup that passed through one of them failed even when a broader namespace above it had a publisher. The lookup now keeps the deepest publisher it passes on the way down instead.

findNode()'s MatchType::Prefix mode existed only to support the old implementation and is now unused; dropped it along with the parameter.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/684)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -881,7 +881,6 @@ MoqxRelay::PublishSetupResult MoqxRelay::publishWithSession(
   auto tracksNode = tracksTree_.findNode(
       pub.fullTrackName.trackNamespace,
       /*createMissingNodes=*/false,
-      NamespaceTree::MatchType::Exact,
       &tracksSessions
   );
   if (tracksNode) {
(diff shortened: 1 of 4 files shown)


mondain (approved):
:lgtm:

> @mondain reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/684#-:-P0ic4yl8JwsFdrevuFy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #685, opened 2026-09-04
fix: prefer exact-track upstream over namespace publisher for FETCH

afrind:
Port of moxygen 50f951ca ("Prefer exact track match over namespace for FETCH"). fetchImpl() resolved the namespace-level publisher first and only fell back to the registry's exact-track entry if that failed. When both existed for the same track, a FETCH was routed to the broader namespace publisher instead of the more specific upstream the relay already had on record for that exact track. Registry lookup now runs first, matching subscribeImpl's resolution order and trackStatusImpl's existing "registry first, namespace tree fallback" pattern.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/685)

### DIFF
--- a/src/MoqxRelay.cpp
+++ b/src/MoqxRelay.cpp
@@ -2590,21 +2590,23 @@ MoqxRelay::fetchImpl(Fetch fetch, std::shared_ptr<FetchConsumer> consumer) {
     }
   }
 
-  auto upstreamPublisher = findUpstreamPublisher(fetch.fullTrackName.trackNamespace);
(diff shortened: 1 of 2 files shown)


mondain (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/685#-:-P0ibs0m2rRfUeLssYgZ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #688, opened 2026-09-05
dev: bisect regressions across moqx, moxygen and the pinned Meta stack

afrind:
A moqx build spans three revision lines — moqx, the moxygen it pins, and the folly/fizz/wangle/mvfst/proxygen/picoquic hashes moxygen pins — so a plain `git bisect` in moqx lands on a `sync: update moxygen` bump and can go no further. scripts/dev/bisect.sh walks all three: phase1 bisects moqx, parking the moxygen checkout on each rev's MOXYGEN_REV itself; phase2 splits a sync bump by building the last-good moqx with its own moxygen source but the first-bad dependency hashes, which says whether the culprit is in moxygen or below it; phase3 bisects inside moxygen by cumulative cherry-pick replay onto a stable base, so mid-merge commits still build and folly does not rebuild at every step. What counts as bad is a predicate, so a crash, hang, test failure or build break bisects as readily as the built-in relay-throughput classifier.

  scripts/dev/bisect.sh phase1 --good v0.3.1 --bad main \
      --test 'ctest --test-dir "$MOQX_BISECT_BUILD_DIR" -R MoQRelay' \
      --test-timeout 120

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/688)

### DIFF
--- a/scripts/dev/bisect.sh
+++ b/scripts/dev/bisect.sh
@@ -1,0 +1,849 @@
+#!/usr/bin/env bash
+# bisect.sh — find the commit that changed moqx's behaviour, across the three
+# repos one moqx build spans: moqx, the moxygen it pins (MOXYGEN_REV in
+# cmake/dependencies.cmake), and the folly/fizz/wangle/mvfst/proxygen/picoquic
+# revisions moxygen pins (build/deps/github_hashes/**/*-rev.txt).
+#
+# WHAT is bisected for is a pluggable predicate — a throughput regression is one
+# use, a crash, hang, test failure or build break are equally good ones.
+#
+# Usage:
+#   bisect.sh phase1  --good REV --bad REV        (--test CMD | --test-perf) [options]
+#   bisect.sh phase2  --good REV --bad REV        (--test CMD | --test-perf) [options]
+#   bisect.sh phase3  --moxygen-good SHA --moxygen-bad SHA
+#                                                 (--test CMD | --test-perf) [options]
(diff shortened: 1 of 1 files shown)


gmarzot:
scripts/dev/bisect.sh:800 — STEP_ARGS never forwards --threshold, so a --test-perf phase1 hits "no threshold set" (exit 128) on its first step and git bisect aborts.

scripts/dev/bisect.sh:675 — cherry-pick --empty=keep needs git 2.45; Ubuntu 22.04 (2.34) and 24.04 (2.43) reject it, so every phase3 replay fails and the search stalls. --keep-redundant-commits is the portable spelling.

scripts/dev/bisect.sh:68 — worth noting in the header that revs before cbeaf418 have no cmake/dependencies.cmake and all skip, so phase1 can't narrow below that commit.

scripts/dev/bisect.sh:311 — if any pinned path is absent at HEAD the whole git checkout HEAD -- fails and the rest stay staged, so the next bisect checkout aborts on a dirty tree. Theoretical today since all three exist across the supported range.

mondain (approved):
:lgtm:

> @mondain reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/688#-:-P1WPh1Y4e94TYx8OX03:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #690, opened 2026-09-05
release: pin MOXYGEN_REV to the moxygen v0.3.3 commit

gmarzot:
One-line release-pin set so the v0.3.3 version-release dispatch can pass validate.

`validate` requires `MOXYGEN_REV` to resolve to the release named in the `moxygen_release` input — the guard that keeps a moqx tag honest about what it consumed. moxygen v0.3.3 tagged `df2a78b6` at 08:03; the daily sync then advanced the pin past it to tip `d425d2f3`, whose only release is a snapshot. Every v0.3.3 dispatch since fails validate identically — the valid window (pin sitting exactly on the release commit) never existed today because the sync jumped straight over it.

This sets the pin to the release commit deliberately. The next daily sync advances it to tip again as normal. `df2a78b6` is an ancestor of the current pin, so this consumes strictly released code — no untested combination.

After merge: dispatch `version release` with `moxygen_release: v0.3.3`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/690)

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -15,7 +15,7 @@
 # on the next reconfigure of an existing build dir. A cached pin would silently
 # shadow the file's value.
 set(MOXYGEN_REPOSITORY "openmoq/moxygen")
-set(MOXYGEN_REV "d425d2f3f6fe4eabbd2737713b1bf410c5b7ccf8")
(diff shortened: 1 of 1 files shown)


gmarzot:
Superseded: releasing moxygen 0.3.4 at tip d425d2f3 instead, which the pin already matches — no pin surgery needed.

PR #691, opened 2026-09-05
release: drive the moxygen release from the moqx dispatch

gmarzot:
Second half of the orchestrated release flow — depends on openmoq/moxygen#411 (the `rev` input and atomic draft publish) landing first.

## What changes

One `version` input now releases both repos. `validate` resolves `MOXYGEN_REV` through `print-release-tag.cmake` exactly as before; the difference is what happens when no **named** (v*) moxygen release points at the pin:

- **Before:** hard fail, with a release window that only existed if the sync happened to leave the pin exactly on a release commit — in practice it required either hand-timing (this morning's 0.3.4 dance) or a manual pin PR.
- **Now:** the run dispatches moxygen's `version release` with the **same version** at `rev = MOXYGEN_REV`, polls the same resolver until the tag appears, and proceeds. Tags agree by construction; the timing race is gone entirely.

The poll is safe against partial releases because of moxygen#411: the moxygen release is a draft until its finalize, and drafts hold no tag ref, so the resolver can only ever see a complete release.

The `moxygen_release` input is **dropped**: it could only restate what the pin already determined or fail — it carried no information (this morning's three consecutive validate failures were exactly this input doing its only possible thing).

If a moxygen release already exists at the pin (any version), it's used as-is with no dispatch — the common case, unchanged.

## Requirements / caveats

- **Merge moxygen#411 first** — the `rev` input and the draft-until-finalize behavior are both load-bearing here.
- **`omoq-sync-bot` needs `actions: write` on openmoq/moxygen** to dispatch the workflow. The token mint is `continue-on-error` and only consulted on the minting path, so if the permission is missing, the common path is unaffected and the minting path fails with an explicit message naming the fix. Worth verifying the app permission before the first orchestrated use.
- Error cases handled explicitly: moxygen `v<version>` already existing at a *different* commit (clear error, no dispatch), and a 60-minute publish timeout (re-dispatching after the moxygen run completes finds the release already minted and takes the common path).

## Testing

YAML parses; the resolver-preference logic (`v*` over `snapshot-*`) matches `MoxygenRelease.cmake`. The dispatch path can only be exercised live; first real use should be the next version release, with the app permission checked beforehand.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/691)

### DIFF
--- a/.github/workflows/version-release.yml
+++ b/.github/workflows/version-release.yml
@@ -23,15 +23,17 @@ name: version release
 #   finalize (--draft=false)
 #
 # Inputs:
-#   version          required. semver, e.g. 1.2.3
-#   moxygen_release  optional cross-check, not a selector. The moxygen consumed
-#                    is always the one MOXYGEN_REV pins — pass the tag you expect
-#                    that pin to resolve to and the run fails early if it doesn't.
+#   version  required. semver, e.g. 1.2.3
 #
-# Behavior matrix, on the moxygen release MOXYGEN_REV resolves to:
-#   tagged v* release (carries portable assets)  → portable matrix runs
-#   rolling snapshot-* release (does not)        → portable skipped (no failure)
-#   no tag at the pin                            → validate fails, nothing tagged
+# The moxygen consumed is always the one MOXYGEN_REV pins. On the release
+# MOXYGEN_REV resolves to:
+#   tagged v* release            → used as-is; portable matrix runs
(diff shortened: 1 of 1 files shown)


gmarzot:
Branch renamed to `ci/orchestrate-moxygen-release` (`release/**` is a publish trigger in this repo, so the old name was misleading and would run ci main on every push). GitHub closed this PR on the rename; continued as #692 with the same content, now referencing openmoq/moxygen#412.

PR #692, opened 2026-09-05
ci: drive the moxygen release from moqx and publish atomically

gmarzot:
Second half of the release flow; needs moxygen#412 first. Two commits.

- **One `version` dispatch releases both repos.** validate resolves `MOXYGEN_REV` as before. If no `v*` moxygen release sits at the pin, it dispatches moxygen's version release with the same version at `rev = MOXYGEN_REV`, watches that run to completion, and re-resolves. A moxygen failure fails this run at once with a link to the moxygen run. If a moxygen release already exists at the pin, nothing is dispatched (the common case). The `moxygen_release` input is dropped: it could only restate what the pin determined, or fail.
- **moqx's own release is published atomically** (second commit, same fixes as moxygen#412): draft until finalize, stale drafts removed first, and portable-build builds the resolved commit instead of the branch tip.

The sync bot's app install already has `actions: write` on moxygen (the upstream sync uses it there). YAML parses; `gh run list --created` and `gh run watch --exit-status` were checked against real moxygen runs. The dispatch path itself can only be exercised live, on the next version release.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/692)

### DIFF
--- a/.github/workflows/version-release.yml
+++ b/.github/workflows/version-release.yml
@@ -23,15 +23,17 @@ name: version release
 #   finalize (--draft=false)
 #
 # Inputs:
-#   version          required. semver, e.g. 1.2.3
-#   moxygen_release  optional cross-check, not a selector. The moxygen consumed
-#                    is always the one MOXYGEN_REV pins — pass the tag you expect
-#                    that pin to resolve to and the run fails early if it doesn't.
+#   version  required. semver, e.g. 1.2.3
 #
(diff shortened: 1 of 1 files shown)


afrind (approved):
itrustyou


> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/692#-:-P1FbY8DDdoMrkW6JyQF:bhb4erc)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna).

PR #693, opened 2026-09-06
ci: publish the release atomically, build the resolved commit

gmarzot:
Ports the second and third parts of openmoq/moxygen#412 to moqx's copy of the version-release workflow. Independent of #692 and mergeable now; #692 only touches `validate`, so it rebases trivially.

## Atomic draft publish

The header comment promised `create-release (--draft)` → `finalize (--draft=false)`, but the create command never passed `--draft`. A moqx release was public and its tag resolvable the moment it was created, while the upload jobs were still running — the same partial-visibility window that produced the half-assembled moxygen v0.3.2. Now the release is a draft until finalize flips it after the upload jobs succeed, so a release either exists complete or not at all. Create also deletes any stale same-tag draft left by a failed prior run (paginated, since snapshot publishes push an old draft off the first page within hours).

`gh release upload` and `gh release edit` find drafts by tag through gh's draft fallback, so the upload jobs need no change. finalize already tolerates a skipped portable-build, so a release without portable assets still gets published.

## Build the resolved commit

`portable-build` checked out the dispatched branch tip, while `promote-snapshot` copies the resolved snapshot commit's assets and `release-image` already builds `snapshot_sha`. A push landing mid-release would ship a portable tarball from a different commit than everything else in the release. It now checks out `snapshot_sha` like release-image does.

## Testing

YAML parses. No workflow in this repo triggers on release events, so moving the "published" moment to finalize changes nothing downstream. Real validation is the next version dispatch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/693)

### DIFF
--- a/.github/workflows/version-release.yml
+++ b/.github/workflows/version-release.yml
@@ -213,14 +213,28 @@ jobs:
             PORTABLE_NOTE="_Portable Linux tarballs not produced for this release: moxygen \`${MOX_REF}\` does not include the required portable assets._"
           fi
 
+          # A failed prior run leaves an invisible draft; a second create with
+          # the same tag_name would not collide (drafts hold no tag ref) and
+          # uploads-by-tag would then be ambiguous. Remove it first. Paginate:
+          # snapshot publishes push an old draft past the first page within hours.
+          gh api --paginate "repos/{owner}/{repo}/releases" \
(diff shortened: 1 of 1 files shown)


gmarzot:
Folded into #692 as its second commit.

PR #697, opened 2026-09-09
relay: (d18+)Reject unsupported Mandatory Track Properties (draft-18 …

akash-a-n:
…§2.5.1)

Add hasUnsupportedMandatoryProperty() (src/relay/TrackProperties.h), a shared helper that flags any extension in the 0x4000-0x7FFF Mandatory Track Property range across both the mutable and immutable extension lists — moqx currently understands none of them.

Wire the check into the three places a peer can hand the relay a mandatory property it can't honor:
- PUBLISH: publish() and publishFromPublisherExec() reject the request with UNSUPPORTED_EXTENSION before installing a forwarder.
- SUBSCRIBE_OK: subscribeUpstreamAndApplyOk() rejects the downstream subscribe if the upstream's OK carries one, instead of applying it to the forwarder.
- FETCH_OK: MoqxCache::fetchUpstream() rejects the fetch and resets the consumer's stream if the upstream's FetchOk carries one, including on the detached background-task path taken when the fast path (live track / known past data) already returned FETCH_OK synchronously.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/697)

### DIFF
--- a/src/MoqxCache.cpp
+++ b/src/MoqxCache.cpp
@@ -9,6 +9,7 @@
 
 #include "MoqxCache.h"
 #include "relay/NullConsumers.h"
+#include "relay/TrackProperties.h"
 #include <folly/logging/xlog.h>
 #include <moxygen/MoQTrackProperties.h>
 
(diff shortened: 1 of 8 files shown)


afrind:
@coderabbitai review

PR #699, opened 2026-09-09
fix: stop treating a partial re-delivered object as a payload mismatch

afrind:
An object large enough to span multiple reads arrives at the relay a chunk at a time. When the cache already held a complete copy, it compared the first chunk's length against the whole object and called it a payload mismatch, resetting the subgroup and tearing down the subscription. The mismatch check now only fires once the re-delivered object is complete again. Replacing a complete cached object with its first chunk also clears the cachedContent marker for that location, so a fetch arriving mid- replacement treats it as a miss and moves past it instead of stalling.

Ported from a fix already merged upstream in facebookexperimental/moxygen that hadn't yet reached moqx's pinned moxygen revision.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/699)

### DIFF
--- a/src/MoqxCache.cpp
+++ b/src/MoqxCache.cpp
@@ -334,7 +334,7 @@ folly::Expected<folly::Unit, MoQPublishError> MoqxCache::CacheGroup::cacheObject
           MoQPublishError(MoQPublishError::MALFORMED_TRACK, "Invalid status change")
       );
     }
-    if (status == ObjectStatus::NORMAL && cachedObject->complete &&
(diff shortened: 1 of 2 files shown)


mondain (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/699#-:-P1WPu2E9phDua9xJfyA:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #708, opened 2026-09-11
Replace relay-hops with MoQ Cluster routing

mondain:
Replace relay-hops with draft-lcurley-moq-cluster-00 routing. Relays negotiate identity and link cost, retain competing namespace advertisements, and select routes by cost and path length while excluding routes that loop through the receiving peer. Advertisement updates stay on their owning stream, and content replacement fences stale subscriptions, fetches, and cache state.

Adds cluster configuration and multiple upstreams with per-link costs, while retaining the legacy single-upstream configuration. Draft 18 peers explicitly negotiate track discovery in both directions. Connection shutdown cancels pending discovery and reconnect work and releases handles on their owning executor. Existing configuration and deployment documentation describes migration and topology requirements.

Depends on https://github.com/openmoq/moxygen/pull/427 and pins its tested commit 8db6696d9577e7f4bce1e038ca707d695e4d8cab. Merge the dependency first.

Validation:
- Default build succeeded; 1,063 tests passed, with 6 existing skips and no failures.
- ASAN/UBSAN subset: 278 passed, with 6 existing skips. Dependencies were not instrumented; this is not full-stack sanitizer coverage.
- Live chain, cycle, pyramid, joining-FETCH, proactive PUBLISH, admin, and shutdown checks passed.
- Formatting and diff whitespace checks passed; public additions were checked for vendor-specific references.

Current boundaries: healthy subscriptions remain pinned when route costs change; cluster FETCH bypasses the shared track-name cache, and route coalescing runs on the relay executor. Anonymous origins do not receive warm-cost updates. Cyclic deployments require draft 18 negotiation and unique nonzero relay IDs; listener version defaults remain unchanged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/708)


peterchave:
In the configuration file the inter relay links are called "upstreams" which made sense in the early implementation, given they are not necessarily hierarchical would it be better to name them something like "peers"

PR #712, opened 2026-09-13
Add a --quic-stack flag to the relay perf harness

afrind:
The perf harness could only drive the mvfst listener, so comparing the two QUIC stacks under the same load meant hand-editing the generated relay config. --quic-stack now selects mvfst or picoquic and defaults to mvfst, so existing invocations are unaffected.

Picoquic refuses the dev cert's `insecure: true`, so the picoquic path serves real TLS: --cert/--key when given, otherwise the repo's localhost test cert, otherwise a throwaway pair minted into the log directory. Congestion control follows the stack, since each spells BBR differently and the relay rejects the other's name.

TEMP (drop before merge): MOXYGEN_REV moves to the head of openmoq/moxygen#430, which makes a picoquic listener bound to "::" reachable from IPv4 peers. It keeps the picoquic paths here on a listener the clients can actually reach. No prebuilt is published for a PR head, so configure.sh falls back to the from-source moxygen superbuild until 430 merges and the pin syncs normally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/712)


gmarzot:
*[`cmake/dependencies.cmake` line 18 at r1](https://reviewable.io/reviews/openmoq/moqx/712#-P1UrigE7phk9S3QaFzG:-P1UrigE7phk9S3QaFzH:b2nhghv) ([raw file](https://github.com/openmoq/moqx/blob/e03bd9eca064c5ffcc292286fd074b601ee760af/cmake/dependencies.cmake#L18)):*
> ```CMake
> # shadow the file's value.
> set(MOXYGEN_REPOSITORY "openmoq/moxygen")
> set(MOXYGEN_REV "61594251884bee03b6bb96e12c524b4b19a8923e")
> ```

Hmm

gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/712)* status: all files reviewed, 1 unresolved discussion (waiting on afrind and michalhosna).

PR #713, opened 2026-09-13
Let the relay topology tests run against the picoquic listener

afrind:
The Python topology harness always configured an mvfst listener, so the relay chain, hops-cycle and pyramid tests only ever exercised one QUIC stack. Setting MOQ_HARNESS_QUIC_STACK=picoquic now points them at the picoquic listener instead; the default is still mvfst, so an unset environment behaves exactly as before.

Picoquic rejects `insecure: true`, so that path also generates a self-signed localhost cert into the test's temporary directory and points the listener at it. This requires openssl on PATH.

TEMP (drop before merge): MOXYGEN_REV moves to the head of openmoq/moxygen#430, which makes a picoquic listener bound to "::" reachable from IPv4 peers. It keeps the picoquic paths here on a listener the clients can actually reach. No prebuilt is published for a PR head, so configure.sh falls back to the from-source moxygen superbuild until 430 merges and the pin syncs normally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/713)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/713)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna).

PR #714, opened 2026-09-13
Run the shell integration tests against both QUIC stacks

afrind:
The shell tests always configured an mvfst listener, so the picoquic listener was only ever exercised by hand. A new test_quic_stack.sh reads MOQ_HARNESS_QUIC_STACK — the same variable the Python topology harness uses — and writes the listener's stack and TLS keys to match, generating a self-signed localhost cert when picoquic is chosen, since picoquic rejects `insecure: true` and serves only exact endpoint paths.

CMake now registers each of these tests once per stack, the picoquic run under a _picoquic suffix. Both variants claim the same ports from test_ports.sh, so a RESOURCE_LOCK keeps ctest from running a pair concurrently. The config-dump test asserts whichever TLS mode its stack actually configured. The qmux test stays single: QMUX is a TCP transport with no QUIC stack to choose.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/714)


afrind:
pico has a shutdown bug here. I will fix it in a subsequent commit that adds multithreading + picoquic, and rebase this commit on top of that, so it will be green.

gmarzot:
a certificate failure doesn't stop the config script (test/make_test_config.sh:46, test/test_admin_cache_purge_race.sh:140). The helper runs inside a heredoc, where set -e doesn't see failures and its exit 1 only leaves the subshell. I ran MOQ_HARNESS_QUIC_STACK=picoquic test/make_test_config.sh 1 2 without --tmpdir. It printed an error but exited 0, and wrote a config with no quic_stack or tls lines. moqx then rejects that config, so the test still fails, but at "relay not ready" rather than at the real cause. test_conformance.sh:102 already does this correctly: store the output in a variable first, then use the variable in the heredoc.

gmarzot (approved):
> @gmarzot reviewed 13 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/714)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on michalhosna).

Issue #696, opened 2026-09-08
Crash on Uni stream before setup complete


peterchave:
Version tested: ghcr.io/openmoq/moqx:v0.3.4-amd64

```
2026-09-08T16:52:17.562036223Z I0908 16:52:17.561866    12 MoqxRelay.cpp:1703] subscribeNamespaceImpl: peer relay detected peer_id=nl-ams-1.moqx.akaleapi.net, reciprocating peer subNs
2026-09-08T16:52:31.209861070Z I0908 16:52:31.209639    12 MoqxRelay.cpp:1703] subscribeNamespaceImpl: peer relay detected peer_id=us-mia-1.moqx.akaleapi.net, reciprocating peer subNs
2026-09-08T16:52:37.691696371Z I0908 16:52:37.690877    13 UpstreamProvider.cpp:350] UpstreamProvider::onMoQSessionClosed error=0 wtError=0
2026-09-08T16:52:47.941642853Z I0908 16:52:47.941388    12 MoqxRelay.cpp:1703] subscribeNamespaceImpl: peer relay detected peer_id=nl-ams-1.moqx.akaleapi.net, reciprocating peer subNs
2026-09-08T16:53:01.356462949Z I0908 16:53:01.356247    12 MoqxRelay.cpp:1703] subscribeNamespaceImpl: peer relay detected peer_id=us-mia-1.moqx.akaleapi.net, reciprocating peer subNs
2026-09-08T16:53:07.929544326Z I0908 16:53:07.928936    13 UpstreamProvider.cpp:350] UpstreamProvider::onMoQSessionClosed error=0 wtError=0
2026-09-08T16:53:18.322509611Z I0908 16:53:18.322281    12 MoqxRelay.cpp:1703] subscribeNamespaceImpl: peer relay detected peer_id=nl-ams-1.moqx.akaleapi.net, reciprocating peer subNs
2026-09-08T16:53:22.125584914Z E0908 16:53:22.125338    13 MoQServer.h:193] proxygen::HTTPException: Got error=268
2026-09-08T16:53:22.259438795Z E0908 16:53:22.259221    13 MoQSession.cpp:6765] Uni stream before setup complete sess=0x737779717010
2026-09-08T16:53:22.260400251Z *** Aborted at 1788886402 (Unix time, try 'date -d @1788886402') ***
[...]

PR #717, opened 2026-09-14
relay: replay retained objects on subscribe; keep join points cached

SimonSl07:
## Summary
A subscriber whose SUBSCRIBE names a location in the past (AbsoluteStart {0,0} — the draft-16
spelling of an unfiltered subscribe, and what `@moq/net` sends by default) is answered
SUBSCRIBE_OK and then nothing. moxygen's forwarder is live-only and clamps the start to
largest+1 (`toSubscribeRange`, MoQLocation.h); that is right for a forwarder with no retention,
but moqx *has* a cache holding exactly the objects asked for and never consulted it for a
SUBSCRIBE. For a catalog track, written once per broadcast, "the next object" is never: only
the very first viewer of a broadcast ever received the catalog, and every later viewer — a
reload, a second tab — hung with no error on either side.

Two defects had to be fixed for this to hold in practice:

1. **Replay from the cache on SUBSCRIBE.** `MoqxCache::replayCachedRange` serves what the
   cache holds in a closed range, never going upstream (unlike `fetch()`, whose miss path
   dereferences its upstream and would turn a subscribe into a FETCH against a publisher that
   may not serve one). `FetchToTrackConsumer` adapts the cache's flat FetchConsumer output onto
   a subscriber's TrackConsumer, one SubgroupConsumer per (group, subgroup), never leaving a
   stream half-open. `MoqxRelay::maybeReplayFromCache` runs it on all three subscribe paths
   that attach to an existing upstream subscription — including the claim path a subscriber
   takes when it lands on an io thread with no local forwarder yet; without that one, the fix
   works or not depending on which thread the viewer lands on. The replay range is
   [start of the largest group, largest]; the live subscription is clamped to largest+1 by
   moxygen, so the two are complementary with no overlap and no gap, and the burst is bounded to one group however much is retained; a hard cap of 256 objects per replay
   backstops that for tracks with very large groups, keeping the newest. The first subscriber (the one whose SUBSCRIBE created
   the upstream subscription) is not replayed to: the publisher answers it directly and a replay
   delivered the catalog twice.

2. **Byte-pressure eviction was deleting the join point.** `evictForByteLimitIfNeeded` walks a
   global group LRU. A catalog is written once at start and media continuously, so its groups are
   the oldest in the cache and are evicted as soon as one 2.5 Mbps stream fills the 16 MB
   default (~51 s) — measured: `catalog groups=0` beside a video track holding 27. After that
   nothing (replay or FETCH) can hand a late viewer the catalog. The newest group of a track
   with a live writeback is now skipped as a victim: one protected group per live track, only
   while it is being published.

## Why not the alternatives
- Making LargestObject inclusive would violate draft-16 §5.1.2 (`{Largest.Group, Largest.Object+1}`).
  Clients that want retained data must ask for it; the relay must then honour the ask.
- Tearing the upstream down when the last subscriber leaves: `checkAndFireOnEmpty` gates on
  `subgroups_.empty()` as well as subscriber count, so a continuously-publishing track never
  reports empty (in 60 min of logs it fired for the audio and catalog tracks and never for video).
  Changing that is a moxygen lifetime change, the failure class behind #696/#676/#663.

## Testing
- Unit tests (`moqx_cache_test`, `moqx_fetch_to_track_consumer_test`): `replayCachedRange` serves a
  closed range in order and ends the fetch once, skips uncached locations, ignores unknown tracks,
  and keeps the newest objects past its cap; byte-pressure eviction spares each live track's
  newest group; `FetchToTrackConsumer` opens one stream per (group, subgroup) and closes every
  stream it opened.
- A draft-16 player that subscribes with AbsoluteStart, reloaded against a live moq-rs
  publisher: previously it timed out waiting for the catalog; now it reconciles the catalog,
  with the relay logging `Cache replay for …/catalog wrote 1 object(s)`. Repeated across io
  threads (MOQX_THREADS=4: replays observed on threads 12, 13, 15).
- `@moq/net` (AbsoluteStart {0,0}) probe joining after the cache had cycled: previously
  `wrote 0` with the catalog track an empty shell; with the eviction change the newest catalog
  group survives and the probe receives a frame.
- No change for LargestObject / NextGroupStart subscribers (spec: after the live edge).

## Log lines added
- INFO `Cache replay for <ftn> wrote N object(s) from {g,o} to {g,o} (asked from {g,o})` —
  including N=0, because when it does not fire the only symptom is a viewer that hangs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/717)


afrind (changes requested):
Per the MOQT spec, you are not supposed to replay cached objects in reply to a SUBSCRIBE.  What you are trying to accomplish is handled via joining FETCH, or in draft-20+ fill-fetch.

https://github.com/moq-wg/moq-transport/pull/1621/files

> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/717#-:-P1Vr4C95_N7WUf2zImX:buf7lvf)* status: 0 of 8 files reviewed, all discussions resolved.

SimonSl07:
Agreed, and thanks for the pointer — §5.1.3 is unambiguous that a past group is joined with SUBSCRIBE (Largest Object) plus a Joining FETCH, and draft-20+ folds the same request into the SUBSCRIBE as FILL_PARAMETERS. I'm withdrawing the replay half.

The eviction change is independent of it and I think still wanted: byte-pressure eviction walks a global LRU, so a track written once (a catalog) is the first victim once media fills the cache, and after that a Joining FETCH for it can't be served from the cache either. I've opened it on its own, on main past the #718 fix: #739. Closing this one in its favour.

PR #721, opened 2026-09-14
Shard the picoquic relay listener across IO threads

afrind:
MoqxPicoRelayServer used to pin every picoquic connection to a single IO thread regardless of the configured thread count, unlike the mvfst and qmux listeners which already spread connections across the whole pool. Switches its base class to moxygen's new MoQPicoQuicShardedServer, which binds one independent picoquic context per IO thread to the same port via SO_REUSEPORT.

Stats collection moves to a per-shard factory since each shard now has its own picoquic context and EventBase instead of one shared pair. That also closes a lifetime hole in the collector: PicoQuicStatsCollector registered itself as a loop observer on the EventBaseStatsCollector and held it as a raw pointer, but never removed itself, so a destroyed collector stayed on the observer list and was called into afterwards. It now takes the EventBase collector as a shared_ptr, keeps a weak_ptr, and deregisters in its destructor while that collector is still alive.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/721)


suhasHere (approved):
:lgtm:

> @suhasHere reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/721#-:-P1Wox2R1D-L1N0aru34:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

suhasHere (approved):
:lgtm:

> @suhasHere made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/721#-:-P1Wp0sT3_1-Bm3l9Zau:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #722, opened 2026-09-14
Wire the relay chain before the forwarder can deliver to it

afrind:
In LocalForwarder mode the passive relay chain hangs off a CrossExecFilter that was attached to the publisher forwarder with a null downstream and only wired to the chain head after the upstream SUBSCRIBE_OK. An upstream that pushes objects in the same batch as its OK reached the filter first, so it opened a subgroup with no downstream, the first object failed, and MoQForwarder dropped the subscriber for the life of the track. The relay then counted no ingest, cached nothing, and read the resulting PublishDone as the track ending. Now the filter is created and wired on relayExec_ before the publisherExec sortie that attaches it, and since getOrCreateFromSubscribe builds the chain head in the same turn, the lookup can no longer miss.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/722)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/722)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #723, opened 2026-09-14
Stop failing local-forwarder subscribers that are waiting their turn

afrind:
A SUBSCRIBE that won the relay-registry first-subscriber race was failed whenever the publisher thread's local-forwarder slot was occupied, even when the occupant was just another SUBSCRIBE still setting up. Failing it tore down the registry entry and threw at everyone parked on it, so the next arrival became the first subscriber and the cycle repeated. Only a ready occupant is a real conflict now, and the pending one is installed over.

That means displacing a pending entry, so its waiters move to the replacement instead of being failed, and markReady, fail and remove all settle the entry before firing the promise so a waiter resuming inline sees the new state. A wake no longer implies the slot is ready, so the subscribe path looks the entry up in a loop rather than once.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/723)


gmarzot (approved):
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/723)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #728, opened 2026-09-17
relay: (d18+)extract PendingRendezvousTree and add rendezvous-timeout…

akash-a-n:
… test coverage

Pull the pending-rendezvous waiter trie out of MoqxRelay into its own PendingRendezvousTree class, along with
the RENDEZVOUS_TIMEOUT clamping logic (takeRendezvousTimeout, kMaxRendezvousTimeout), which has no dependency on relay state and now lives alongside the tree it feeds.

Add missing test coverage:
- RendezvousTimeoutClampedToMax: 30s clamp on oversized client timeouts
- TimedOutWaiterPrunesTree: trie node is pruned after a waiter times out
- CancellationWhileParkedStopsWithoutResult: co_stopped_may_throw path when a parked SUBSCRIBE is cancelled

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/728)


afrind (approved):
This is fine and thanks for hanging with it.  I rebased everyone on main which should bring tests green.  I'm going to meat proxy some AI nits, you can address or push back/defer if you want.

  Correctness
  
  One thing I would fix before landing:                              

  A test comment asserts the opposite of what the code does — test/MoqxRelayRendezvousTests.cpp:470-474. It
  says moqx "does not" route to a root publisher because findNode() "refuses to prefix-match at the root
  (nodePtr.get() != &root_)", then the test asserts hasValue(). NamespaceTree::findPublisherSession seeds
  deepestPublisher from root_.publisherSession_ (src/NamespaceTree.cpp:82), and that guard does not appear in
  the function at all. It landed in 505472bd (#684), before this stack's base — so the comment describes a
  bug that was already gone when it was written, and will send the next reader hunting for a defect that does
  not exist.

  Verified sound, for the record:

  - The reentrancy fix holds. Both wake paths complete all mutation and ancestor pruning before signalling; a
    waiter resumed inline finds a pruned path and its eraseWaiter no-ops.
  - Both wakes run on relayExec_ in all three modes, and doPublishNamespace's wake sits after the
    allowed-prefix and setPublisher checks, so a rejected announce does not wake anyone.
  - timekeeper_ is declared after ownedRelayExec_, so it is destroyed before the EventBase it references.
  - The param-erase union is equivalent to moxygen's. takeRendezvousTimeout consumes 0x04 when downstream is
    v18+, makeUpstreamSubReq's isV18Plus(upstreamSession) when upstream is — together that is moxygen's
    downstreamIsV18Plus || upstreamIsV18Plus.

  Worth a deliberate decision rather than a fix:

  - One park, never re-armed. A wake returns nullopt and the caller's ordinary path reports DOES_NOT_EXIST,
    so a publisher that announces and withdraws — or an upstream SUBSCRIBE that fails — consumes the entire
    rendezvous timeout instead of leaving the subscriber waiting out its clock. This matches moxygen exactly
    (it re-checks and returns "publisher no longer available"), so it is parity, not a regression. If a
    follow-up is wanted anywhere, it is here, not in the LF path.
  - kMaxRendezvousTimeout is hardcoded at 30s (src/PendingRendezvousTree.h:86). moqx has a config system;
    this is the kind of ceiling an operator will want to move.
  - No relay-level cap on parked waiters. Each is a suspended coroutine frame, a TimedBaton and a trie node
    held for up to 30s, bounded only by whatever MAX_REQUEST_ID moxygen grants per session. I did not find a
    moqx-level limit.

  Simplification

  eraseFromNode (src/PendingRendezvousTree.cpp:47, ~30 lines, recursive) is applyAtNamespaceNode with a
  different lambda — walk down, act on the node, prune empty ancestors on the way up. The two are
  behaviourally identical, including at the root. Collapsing it leaves one walk primitive and makes
  eraseWaiter and wakeForTrack the same shape.

  isV18Plus in MoqxRelay.cpp duplicates the version gate inside takeRendezvousTimeout; one exported helper
  covers both.

  The class itself is a better shape than the equivalent code sitting inline in moxygen's MoQRelay — a
  candidate to push upstream and delete one of the two copies.

  Performance

  Fine. On an empty trie a wake costs one F14 miss and allocates nothing, because the walk breaks at depth 0
  before path is touched — so the common PUBLISH path is untaxed. The one nit: when the walk does descend,
  path.emplace_back(node, part) (src/PendingRendezvousTree.cpp:108) copies a std::string per namespace level.
  string_view or an index would drop that.

  Test coverage

  Good on the core mechanics — park, wake-by-track, wake-by-namespace, ancestor wake, sibling isolation,
  timeout, cancellation, trie pruning, and all three relay modes. Four gaps:

  1. isV18Plus(upstreamSession) has zero coverage. Every pubSession in the file is createMockSession()
  3. The descendant-announce case is missing from PendingRendezvousTreeTest: a waiter parked at a must stay
     parked when wakeUnderNamespace({"a","b"}) fires, since a publisher at a/b cannot serve a/track. There is
     a sibling test but not this one.
  4. The spurious-wake path is untested — announce, then withdraw before the waiter resumes.

> @afrind reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/728#-:-P1v00WED0wvywMz0vQs:b-mw7jmh)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on akash-a-n).

PR #734, opened 2026-09-18
cache: stop a split object being cached and marked non-existent at once

afrind:
An object delivered in pieces completes in objectPayload, which stepped the fetch cursor only when the piece carried the fetch's fin. With another object behind it the cursor stayed parked on the completed object, so that object's markNonExistentTo() -- which records the positions a fetch skipped as non-existent -- recorded it as a gap. The object was then in the cache and in gaps at the same time, and every later fetch that hit the cache skipped it, delivering the next object ID in its place.

Step past the object as soon as it completes, as object() already does for one that arrives whole, leaving fin to do only the tail-marking.

This is why the six cached-FETCH conformance tests failed while the two uncached ones passed: the first fetch still forwards the split object live, so only a later fetch served from the poisoned cache entry shows it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/734)


afrind:
@coderabbitai review

akash-a-n (approved):
:lgtm:

> @akash-a-n made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/734#-:-P2DS3hy93OqVOKKKmaH:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved.

PR #735, opened 2026-09-18
cache: stop assuming a fetched object has a payload

afrind:
Draft 16 removed the Object Status field from Objects delivered via FETCH, so a zero-length object now reaches the cache from the codec with a null payload. That is the ordinary encoding for an empty object rather than something only a misbehaving peer sends. Two payload clones — one serving a cached object, one writing an upstream fetch back — assumed a payload was always present and crashed on it.

Port of the MoQCache portion of moxygen fb4f0bd6; the framer and session changes come with the pin.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/735)


afrind:
@coderabbitai review

akash-a-n (approved):
This and https://reviewable.io/reviews/openmoq/moqx/750 are duplicates of each other. I'll let you decide which one to keep.

:lgtm:

> @akash-a-n reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/735#-:-P2DBcGPELtKs4jRDja4:bimkqit)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on afrind).

PR #737, opened 2026-09-18
Keep reconnecting when an upstream drops mid-handshake

afrind:
An upstream session that dies between CLIENT_SETUP and SERVER_SETUP reaches the client as folly::OperationCancelled, because MoQSession cancels its own token when it cleans up. UpstreamProvider's reconnect loop read any cancellation as a stop request and exited for good, so the relay stayed disconnected from its upstream with no retry. The loop now treats cancellation like any other connect failure and only exits when stop() was called. This is what failed relay_pyramid_picoquic under asan, where relay D never learned the namespace it was waiting on.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/737)


afrind:
@coderabbitai review

akash-a-n (approved):
:lgtm:

> @akash-a-n made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/737#-:-P2DUlK-0JZDUnsLs08m:bnfp4nl)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on afrind).

___
*[`src/UpstreamProvider.cpp` line 111 at r1](https://reviewable.io/reviews/openmoq/moqx/737#-P2DUNddEEakhItDEpHX:-P2DUNddEEakhItDEpHY:bh2pdxk) ([raw file](https://github.com/openmoq/moqx/blob/16b47de6b4df9d6aeff9f9e84087157bf1482d90/src/UpstreamProvider.cpp#L111)):*
> ```cpp
> static constexpr auto kInitialReconnectBackoff = std::chrono::seconds(1);
> static constexpr auto kMaxReconnectBackoff = std::chrono::seconds(60);
> ```

Do we not want a max number of retries? Say if the upstream is down?

afrind (commented):
> @afrind made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/737)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on akash-a-n).

___
*[`src/UpstreamProvider.cpp` line 111 at r1](https://reviewable.io/reviews/openmoq/moqx/737#-P2DUNddEEakhItDEpHX:-P2Ddg3E2gT2xA3kCS8-:b-fxruhr) ([raw file](https://github.com/openmoq/moqx/blob/16b47de6b4df9d6aeff9f9e84087157bf1482d90/src/UpstreamProvider.cpp#L111)):*

Do we not want a max number of retries? Say if the upstream is down?
</blockquote></details>

Right now it will backoff up to 60s and then keep trying once/min to reconnect

Issue #718, opened 2026-09-14
Cache eviction triggers 100% CPU


peterchave:
Core located on /root/core.1182924 on us-ord-2 test machine if needed.

Inspection of threads using gdb:

```
Thread 8 (LWP 1183004):
#0  0x00007b5455535e9d in writev () from ./container_sysroot/lib/x86_64-linux-gnu/libc.so.6
#1  0x000056dafdc7e733 in long folly::fileutil_detail::wrapvFull<long (*)(int, iovec const*, int)>(long (*)(int, iovec const*, int), int, iovec*, int) ()
#2  0x000056dafdbd2995 in folly::AsyncFileWriter::writeToFile(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, unsigned long) ()
#3  0x000056dafdbd2a4e in folly::AsyncFileWriter::performIO(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, unsigned long) ()
#4  0x000056dafdbd372a in folly::AsyncLogWriter::ioThread() ()
#5  0x00007b54557f04a3 in ?? () from ./container_sysroot/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007b54554c11f5 in ?? () from ./container_sysroot/lib/x86_64-linux-gnu/libc.so.6
#7  0x00007b5455540b50 in clone () from ./container_sysroot/lib/x86_64-linux-gnu/libc.so.6

Thread 7 (LWP 1183005):
#0  0x00007fff45bdd763 in ?? ()
#1  0x00007fff45bddcf0 in ?? ()
#2  0x00007b5455507439 in clock_gettime () from ./container_sysroot/lib/x86_64-linux-gnu/libc.so.6
#3  0x00007b54557e7192 in std::chrono::_V2::system_clock::now() () from ./container_sysroot/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x000056dafdbe966d in folly::LogMessage::LogMessage(folly::LogCategory const*, folly::LogLevel, folly::Range<char const*>, unsigned int, folly::Range<char const*>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&&) ()
[...]

PR #739, opened 2026-09-18
cache: never evict a live track's newest group for byte pressure

SimonSl07:
## Summary
Byte-pressure eviction walks the global group LRU oldest-first, so a track written once at the start of a broadcast — a catalog — is the first thing evicted once a media track is writing beside it. Measured on a live relay: `catalog groups=0` (last written 590 s earlier) beside a video track holding 27 groups, within a minute of one 2.5 Mbps stream filling the 16 MB default. From then on a late subscriber's Joining FETCH for that group misses the cache and is forwarded upstream, and the live subscription never delivers it again.

The newest group of every track with a live writeback is skipped when choosing a victim. That protects at most one group per live track, only while it is being published; afterwards the group is an ordinary LRU entry again.

Split out of #717 following review: the replay-on-SUBSCRIBE half is withdrawn, and this is the part that makes the Joining FETCH answerable from the cache.

## Notes
- Based on main `8fab4565`, which carries the #718 fix. The walk is a single pass carrying its iterator: every step erases the entry or moves past it, so it cannot spin, and nothing is re-walked after an eviction.
- The byte cap can now be exceeded by up to one group per live track. Open subgroups already pinned groups past it; this adds one bounded case.
- `TestTrackEvictionWithOpenSubgroupDoesNotSpin` now expects 200 bytes: its second track is live with one group, so that group is a join point and stays.

## Testing
Two unit tests in `moqx_cache_test`:
- `ByteLimitEvictionSparesEachLiveTracksNewestGroup`: a once-written six-group track beside a track that keeps writing past the byte limit. Older groups of both are evicted, each live track keeps its newest, and a FETCH for the surviving catalog group is served from the cache with no upstream call (`upstream_` is a StrictMock).
- `ByteLimitEvictionTakesTheNewestGroupOnceTheTrackIsNotLive`: the same with the once-written track's writeback released first; its newest group is then evicted like any other.

The first revision of this change ran on a live relay with the cache at its byte cap: the catalog track kept `groups=1` beside the media track where before it read `groups=0` within a minute. The loop was rewritten since; CI is this revision's build.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/739)


afrind:
The concept here seems like a good one, I think.  Looks like there's a failing test to address.  I'm not sure why this isn't showing in our review tool (reviewable, cc: @michalhosna).

Also, please make a comment and commit summary trim pass.

afrind (changes requested):
Please fix the broken test

afrind:
My agent came back with some review comments as well:

1) the cache is no longer bound by max cache bytes.  I think this was already true though, since there was already a concept of pinning things being written to, I think I can live with it, but it's something we should surface via metrics, or possibly bound the amount of time the pin holds without anyone reading from it, at least by MAX_CACHE_DURATION?

2) turns out globalGroupLRU is actually a FIFO, but we should fix that separately

3) the eviction loop isn't very efficient: it restarts at rbegin after every eviction, likely expending duplicate work that could be saved

4) the agent flagged that if objects from the largest group get evicted at some point for another reason before the track advances (purge?) then it can't be recovered.

5) The walk already resolved trackIt for each entry and nothing between the break and here touches cache_, so the second find() is a wasted hash lookup on a path that runs per object and per payload chunk. Same 
story with `const auto [ftn, groupID] = *victim`: that copies a whole FullTrackName (namespace vector plus  the name string) on every eviction, where the old code just used the stable map key.

SimonSl07:
Pushed a revised commit (force-push, still one) with comments and message trimmed.

**Failing test:** `TestTrackEvictionWithOpenSubgroupDoesNotSpin` is the new rule working, not the walk misbehaving — its `track2` is live with one group, so that group is now a join point and survives where it used to be the oldest entry. The test expects 200 bytes and asserts the group is kept; the termination check is unchanged.

On the agent's points:

1. Agreed the cap is soft by up to one group per live track, on top of the open-subgroup pins that already exceeded it. I can add a gauge for join-point bytes as a follow-up. A time bound via `MAX_CACHE_DURATION` reinstates the bug for the case this fixes: the catalog is written once and read by every joiner for the life of the broadcast.
2. Agreed, and it cuts the same way: under a FIFO the catalog sits at the tail however often it is read.
3. Fixed — single pass carrying the iterator, nothing re-walked after an eviction.
4. Right, this covers byte pressure only. Purge is deliberate and the track limit only takes non-live tracks, but per-object duration expiry in `getCachedObjectMaybe()` still drops a catalog object once it is older than the relay default or the publisher's per-track duration. Same bug on another path; I can follow up exempting the join point there if you agree with the shape.
5. Fixed — `trackIt` from the walk is reused and only the group id is copied.

One limitation: `largestGroupAndObject` is monotonic for the `CacheTrack`'s life, so a track re-published under the same name with restarted group numbering, while the old run is still cached, protects the old run's group rather than the new one's.

SimonSl07:
Ran a pass to trim the comments and commit summary and also to better match repo code style, should fit better now.

afrind:
> A time bound via MAX_CACHE_DURATION reinstates the bug for the case this fixes: the catalog is written once and read by every joiner for the life of the broadcast.

Yes there's a tension here, but I don't think we can sustain "live groups are never evicted" across all tracks without bounds - it opens up a DOS window.  Memory is finite and groups are not.  MAX_CACHE_DURATION (and the cache's own config that limits it) is already a publisher's signal about the useful caching lifetime of objects, so that needs to be respected regardless.

A cache miss is not a death sentence.  The relay will just go back to the original publisher and ask for the catalog again.  Surely the OP can still produce it, since there's always a chance that the relay restarts and the in memory cache is lost.

> Purge is deliberate and the track limit only takes non-live tracks, but per-object duration expiry in getCachedObjectMaybe() still drops a catalog object once it is older than the relay default or the publisher's per-track duration. Same bug on another path

See above, these are features and important defensive measures, not bugs.

> One limitation: largestGroupAndObject is monotonic for the CacheTrack's life, so a track re-published under the same name with restarted group numbering, while the old run is still cached, protects the old run's group rather than the new one's.

You might need your agent to re-read the moq-transport spec.  You cannot use the same name and restart the group numbering -- at least you cannot republish the same track namespace + name + group ID + object ID with different content.  That violates the MOQT object model and constitutes a malformed track.

afrind (approved):
> @afrind reviewed 3 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/739)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SimonSl07).

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/openmoq/moqx/739#-P1uAzkPAfATC5k9pGrE:-P1uAzkPAfATC5k9pGrF:b6cecvn):*
I don't think this last sentence is an accurate description.  A live subscription never delivers cached data.  Further, if the catalog is evicted and re-fetched, it will be cached again and granted a new life through the LRU. 
___
*[`src/MoqxCache.cpp` line 2459 at r1](https://reviewable.io/reviews/openmoq/moqx/739#-P1uBuHA5ZEXbNv7J-3t:-P1uBuHA5ZEXbNv7J-3u:b-5wyoq9) ([raw file](https://github.com/openmoq/moqx/blob/d7c34251e8b1d621000841479f99760437a5adc4/src/MoqxCache.cpp#L2459)):*
> ```cpp
>   // every step erases the entry or moves past it, so the loop cannot spin.
>   auto it = globalGroupLRU_.end();
>   while (totalCachedBytes_ > targetBytes && it != globalGroupLRU_.begin()) {
> ```

There's a pathological case here which is that the cache is over the limit with a large number of tracks that are not publishing anything.  Every new object walks the entire cache only to evict nothing.  I filed #741.

Protecting the relay's memory is more important than cache hit rate.

PR #742, opened 2026-09-20
Ask upstream for objects only when a subscriber is forwarding

afrind:
publishWithSession counted every matching namespace subscriber toward shouldForward, ignoring each one's forward flag, so a track whose subscribers all had forward=0 was answered with PUBLISH_OK(forward=true). The publisher then sent objects the forwarder had no forwarding subscriber to deliver, and nothing corrected it because forwardChanged fires only when forwardingSubscribers_ leaves 0, which never happened. Count only subscribers that are forwarding.

- - -
This change is [Reviewable](https://reviewable.io/reviews/openmoq/moqx/742)


akash-a-n (approved):
:lgtm:

> @akash-a-n reviewed 2 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/openmoq/moqx/742#-:-P2DbyDQ1SoQeVC7k9_A:bnfp4nl)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on afrind).

